PR #791, opened 2015-01-07
Wowza keeps reinstalling after machine reboot

ppp0:
Follow-up of #667 

Happens after rebooting a machine.

Excerpt from `syslog` after booting:

```
2015-01-07T07:32:12.295620+00:00 wowza2 puppet-agent[3538]: Reopening log files
2015-01-07T07:32:12.421481+00:00 wowza2 puppet-agent[3538]: Starting Puppet client version 3.7.3
2015-01-07T07:32:18.502455+00:00 wowza2 kernel: [   23.858911] bond0: no IPv6 routers present
2015-01-07T07:32:19.622422+00:00 wowza2 kernel: [   24.976720] bond1: no IPv6 routers present
2015-01-07T12:32:26.867502+00:00 wowza2 ntpdate[2259]: step time server 132.163.4.102 offset 18000.497650 sec
2015-01-07T12:32:47.243665+00:00 wowza2 ntpdate[4445]: step time server 132.163.4.102 offset -0.000387 sec
2015-01-07T12:32:56.688632+00:00 wowza2 monit[5072]: monit: Status not available -- the monit daemon is not running
2015-01-07T12:32:58.916585+00:00 wowza2 monit[5095]: monit: Status not available -- the monit daemon is not running
2015-01-07T12:33:04.596379+00:00 wowza2 monit[5136]: Starting monit daemon with http interface at [*:2812]
2015-01-07T12:33:04.605071+00:00 wowza2 monit[5138]: Starting monit HTTP server at [*:2812]
2015-01-07T12:33:04.605110+00:00 wowza2 monit[5138]: monit HTTP server started
2015-01-07T12:33:04.605145+00:00 wowza2 monit[5138]: 'base' Monit started
2015-01-07T12:33:04.784132+00:00 wowza2 monit[5138]: 'wowza' process is not running
2015-01-07T12:33:04.860845+00:00 wowza2 monit[5138]: 'wowza' trying to restart
2015-01-07T12:33:04.860858+00:00 wowza2 monit[5138]: 'wowza' start: /etc/init.d/wowza
2015-01-07T12:33:04.938909+00:00 wowza2 monit[5138]: 'ntp' process is not running
2015-01-07T12:33:04.981225+00:00 wowza2 monit[5138]: 'ntp' trying to restart
2015-01-07T12:33:04.981237+00:00 wowza2 monit[5138]: 'ntp' start: /etc/init.d/ntp
[...]

PR #792, opened 2015-01-07
Change runlevel to start puppet daemon

ppp0:
Resulting from insights gained while investigating #791 

`puppet` should start a bit later (runlevel 3 means networking has been started)

@tomaszdurka @njam @kris-lab opinions? agree? thoughts?

### DIFF
--- a/modules/bipbip/templates/init.sh
+++ b/modules/bipbip/templates/init.sh
@@ -4,7 +4,7 @@
 # Provides:				bipbip
 # Required-Start:		$syslog
 # Required-Stop:		$syslog
-# Default-Start:		2 3 4 5
+# Default-Start:		3 4 5
 # Default-Stop:			0 1 6
 # Short-Description:	Bipbip
 # Description:			Gather services data and store in CopperEgg
--- a/modules/copperegg_revealcloud/templates/init.sh
+++ b/modules/copperegg_revealcloud/templates/init.sh
@@ -5,7 +5,7 @@
 # Provides:				revealcloud
 # Required-Start:		$syslog
 # Required-Stop:		$syslog
-# Default-Start:		2 3 4 5
+# Default-Start:		3 4 5
 # Default-Stop:			0 1 6
 # Short-Description:	CopperEgg RevealCloud collector
 # Description:			CopperEgg RevealCloud collector
--- a/modules/mms/templates/init
+++ b/modules/mms/templates/init
@@ -4,7 +4,7 @@
 # Provides:						<%= @agent_name %>
 # Required-Start:			$syslog
 # Required-Stop:			$syslog
-# Default-Start:			2 3 4 5
+# Default-Start:			3 4 5
 # Default-Stop:				0 1 6
 # Short-Description:	<%= @agent_name %>
 # Description:				Get MongoDB cluster status and sens to MMS service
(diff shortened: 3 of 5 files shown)


ppp0:
@njam please review

njam:
Hm, all other init script have `2 3 4 5` too.
Change it there as well?

njam:
..for this change to apply, do we need to reinstall all init scripts?

ppp0:
You're right. All other scripts have similar settings. We should investigate if a change is needed, though

It is certainly needed for `puppet` (Reason for this PR).

To reinstall the script.. IMO `puppet` should pick up the change in file's MD5 and restart the daemon?

```pp
[..]
   '/etc/init.d/puppet':
      ensure => file,
      content => template("${module_name}/agent/init"),
      group => '0',
      owner => '0',
      mode => '0755',
      notify => Service['puppet'];

[..]
```

@kris-lab wdyt?

kris-lab:
looks good to me; I think the daemon will be restarted after change.

ppp0:
uh, sorry, I'm being reminded of `update-rc.d` - we obviously need to run this one to properly install init scripts (ie which level they get run/stopped on at boot/shutdown)

ppp0:
@njam @kris-lab I revised the PR and changed the default runlevel to some more scripts. IMO it makes sense to wait until networking is enabled (runlevel 3) to start them. Please review

njam:
Why is this change needed in the first place?
Isn't the cause for our problems the environment variable `COLUMNS=80`?

Also I can't find documentation on that runlevel 3 means "start networking" - did you test this manually? or did you find some docu on it?
-> https://www.debian.org/doc/manuals/debian-reference/ch03.en.html#_stage_4_the_normal_debian_system

ppp0:
I do not agree:
Clearly, `COLUMNS` in the environment affects `dpkg -l`, but there's more env variables (`runlevel` being the most prominent) which do not belong there in normal operation.

To get information about runlevels I suggest to consult http://en.wikipedia.org/wiki/Runlevel

kris-lab:
Run levels docu is [here](http://www.tldp.org/LDP/sag/html/run-levels-intro.html)

`3` means networking for multiuser
`2` there is only local user

njam:
Yes, but both docus you've shared are *generic linux* (unclear what it means exactly?). I can't find any documentation for *Debian* that says anything more than that runlevels 2-5 are "multi user".

Diff of rc1 and rc2 on a fresh Debian box:
```
## diff <(ls -1 /etc/rc1.d) <(ls -1 /etc/rc2.d)
1,9d0
< K01atd
< K01exim4
< K01polipo
< K01puppet
< K01vboxadd-service
< K02vboxadd
< K03rsyslog
< K05nfs-common
< K05rpcbind
11d1
< S01killprocs
12a3,17
> S01vboxadd
> S02vboxadd-service
> S13rpcbind
> S14nfs-common
> S16polipo
> S16rsyslog
> S16sudo
> S17acpid
> S17atd
> S17cron
> S17exim4
> S17ntp
> S17puppet
> S17rsync
> S17ssh
14c19,20
< S20single
---
> S20rc.local
> S20rmnologin
```

and diff of rc2 with rc{3,4,5}:
```
# for i in $(seq 3 5); do diff <(ls -1 /etc/rc2.d) <(ls -1 /etc/rc${i}.d); done
#
```

and after installing both mysql and apache2 servers that's still the case:
```
# for i in $(seq 3 5); do diff <(ls -1 /etc/rc2.d) <(ls -1 /etc/rc${i}.d); done
#
```

So it looks to me like Debian treats runlevels 2-5 all the same, and starts all daemons in run level 2?

ppp0:
Closing after discussion (original problem solved, Debian without noticeable difference between runlevels 2 and 3)

PR #795, opened 2015-01-09
Make puppet $vardir world readable

kris-lab:


kris-lab:
@njam please have a look; this is the easiest working way!

njam:
Add an assertion to the agent spec, with a note that we want `/var/lib/puppet/state` to be world-readable?

njam:
Owner should be puppet

kris-lab:
@njam please review

njam:
lgtm

PR #798, opened 2015-01-09
Add tests with multiple mysql users

njam:
@kris-lab please review

Had these lying around on my index. I think I wrote them once because we had problems when defining multiple `mysql:user` resources.
I'd suggest to add them. wdyt?


kris-lab:
yep sure! looks good

PR #805, opened 2015-01-15
Detect changes from origin/<branch>

njam:
@kris-lab please review

The local `master` is not always fast-forwarded

### DIFF
--- a/build-ci.sh
+++ b/build-ci.sh
@@ -13,5 +13,5 @@ if [ -z "${ghprbTargetBranch}" ]; then
     bundle exec rake test
 else
     # Pull request build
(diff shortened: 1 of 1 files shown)


njam:
...something still doesn't work with `diff`, checking...

kris-lab:
lgtm what is there now! 

what you mean; is it still not working?

njam:
Yup somehow `origin` was not fetched - I'm doing it manually now - please check!

kris-lab:
hmm makes sense; so the problem was that changes were not fetched? or also new branches?

kris-lab:
lgtm btw

njam:
`origin` was not fetched, so the `diff ^origin/master` would show changes from other PRs..

PR #811, opened 2015-01-16
Install latest bipbip version

njam:


njam:
Paired with @kris-lab

PR #814, opened 2015-01-17
Adjust NPM download URL, make "curl" follow redirects

njam:


njam:
Self-merging

@kris-lab @ppp0 please review

PR #816, opened 2015-01-17
Destroy VM after build

njam:
This reverts commit b1c7eb74cbadcdd5466f225e2df247d8fca6efa1.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -19,7 +19,7 @@ end
 
 namespace :test do
   task :cleanup do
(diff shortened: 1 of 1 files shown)


njam:
Self merging

@kris-lab only "halting" the box is not a good idea for two reasons:
- Jenkins cleans out the *workspace* folder, so `.vagrant` is lost and vagrant loses control over the previous VM
- We should regularly rebuild the VM anyway for some operations to happen (e.g. `apt-get update`).

I'm reverting for now..

PR #817, opened 2015-01-17
Use puppet-syntax gem instead of "puppet parser validate"

njam:
@kris-lab @ppp0 please review

This does `puppet parser validate` plus additional checks.
See https://github.com/gds-operations/puppet-syntax


kris-lab:
lvgtm. like it!

njam:
What's cool is they also do syntax check for YAML and .erb!

PR #815, opened 2015-01-17
New deb-multimedia key

njam:

PR #819, opened 2015-01-17
Use short gpg key fingerprint

njam:
Followup of #815

Need to use short key fingerprint, otherwise it will not match the output of `apt-key list` and install over and over again..


njam:
Self-merging

@kris-lab please check

njam:
@ppp0 also fyi!

PR #822, opened 2015-01-18
Move packages from cargomedia/puppet-cm repo

tomaszdurka:


tomaszdurka:
@njam please review

njam:
retest this please

njam:
Looks good!

(Some specs would be good, but we can add them later)

tomaszdurka:
Thought the same on the other hand it's mostly set of other packages. Maybe except stream and webserver.

I will add some later on.

PR #830, opened 2015-01-20
Make composer::project target non-optional

tomaszdurka:


tomaszdurka:
Hi @ppp0. I talked with Reto and we decided that it's better to not have default target location for packages. Whenver you want to install one you need to specify `$target` param.

ppp0:
Makes sense in the end.. was it that we didn't want to overload the define with too many arguments?

Ah, LGTM

tomaszdurka:
Resaon was to simplify that, but it was bringing more confusion in the end. I thnk it's cleaner in usages like s3export.

PR #813, opened 2015-01-20
/var/log/syslog should be world-readable

ppp0:
We set in `rsyslog.conf`:
```
$FileCreateMode 0644
```

But still on some hosts this doesn't apply:
```
root@rep1-db3.mongodb.example.com:~$ ls -l /var/log/syslog
-rw-r----- 1 root adm 36012 Jan 16 19:06 /var/log/syslog
```

@ppp0 could you have a look? Is it only applied when the file is created? Maybe it's easiest to manage the file with puppet?
cc @kris-lab


ppp0:
@njam fancy reviewing? thx

njam:
lvgtm!

PR #832, opened 2015-01-21
Refactor service installation to sysvinit::script define

ppp0:


ppp0:
While working on https://github.com/cargomedia/puppet-packages/commit/9f076deebf7432dbabf03687d4f76ce114d2d1b5 I noticed we might need to streamline this everywhere

@njam please review @kris-lab fyi

njam:
I don't get it, this should have two effects, right?
- Ordering (apply `helper::service` after `File`): Doesn't make sense because the file is defined *in* the `helper::service`?
- Notification (refresh the `helper::service` when `File` changes): Again same as a above. Also this could be done in the `helper::service` directly?

Can you elaborate?

ppp0:
* https://github.com/cargomedia/puppet-packages/blob/master/modules/helper/manifests/service.pp - " file is defined in the helper::service?" ?? Do I miss something?
* We bumped into this recently while [discussing modifying the default runlevel](https://github.com/cargomedia/puppet-packages/pull/792) - To apply changes one needs to re-run `update-rc-d`?

njam:
Ok I think I understand now: The goal is to reinstall the init script when it changes, to reflect possible changes of runlevels?

- Is `update-rc.d <daemon> defaults` sufficient then, no need to uninstall first?
- The `unless` of the `Exec` is looking in `/etc/rc3.d/` - dangerous - can we use [`invoke-rc.d --query`](https://github.com/puppetlabs/puppet/blob/cb4ad1f19b043a70ba17e4103a25f5c97a76948b/lib/puppet/provider/service/debian.rb#L34)?
- Can't we move the `subscribe` *into* the `helper::service` define (for the first `exec`)

ppp0:
Makes sense!

ppp0:
May be make starting the service optional/configurable?

njam:
> May be make starting the service optional/configurable?

I would always start it, as it's debian defacto standard.
*If* we have a different use case, then we can revisit?

ppp0:
@njam @kris-lab who wants to review?

njam:
Could you prepare a separate formatting pull-request with only code styles applied? It will make reviewing easier.

modules/helper/manifests/service.pp
```diff
@@ -1,8 +1,19 @@
...
+    content => $init_file_content,
+    owner   => '0',
+    group   => '0',
+    mode    => '0755',
+  }
+  ~>
 
   exec { "update-rc.d ${name} defaults":
-    path   => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
-    unless => "ls /etc/rc3.d/ | grep -- '^S..${name}$'",
```
ppp0: Explanation: Apparently, `defaults` always installs the script in start levels 2,3,4,5, stop 0,1,6 regardless of what's in the comment section of the init file

Excerpt fomr Wheezy's `usr/sbin/update-rc.d`:
```pl
# Process the arguments after the "defaults" keyword.

sub defaults {
    my @argv = @_;
    my ($start, $stop) = (20, 20);

    &usage ("defaults takes only one or two codenumbers") if ($#argv > 2);
    $start = $stop = $argv[1] if ($#argv >= 1);
    $stop  =         $argv[2] if ($#argv >= 2);
    &usage ("codenumber must be a number between 0 and 99")
        if ($start !~ /^\d\d?$/ || $stop  !~ /^\d\d?$/);

    $start = sprintf("%02d", $start);
    $stop  = sprintf("%02d", $stop);

    $stoplinks[$_]  = "K$stop"  for (0, 1, 6);
    $startlinks[$_] = "S$start" for (2, 3, 4, 5);

    1;
}
```

Executing `update-rc.d` this way is therefore idempotent and it can be exec's on refresh only

ppp0: Is it needed at all to execute it on refresh? Not so convinced anymore.. May be on init file creation is enough?

ppp0:
files left out to refactor (to be discussed):

Not installed in a `rcX.d/` tree
* https://github.com/cargomedia/puppet-packages/blob/master/modules/monit/manifests/init.pp#L80 - This one does not need to be `update-rc.d`-installed as is run on boot due to https://github.com/cargomedia/puppet-packages/blob/master/modules/monit/templates/default

Non-trivial orchestration, needs careful review by @kris-lab :
* https://github.com/cargomedia/puppet-packages/blob/master/modules/mongodb/manifests/core/mongod.pp
* https://github.com/cargomedia/puppet-packages/blob/master/modules/mongodb/manifests/core/mongos.pp

ppp0:
@kris-lab @njam OK - now it's easier to review (styling merged upstream)

who dares?

kris-lab:
@ppp0 this refactoring (exec -> helper::service)  is not going to restart services after apply, yes?

ppp0:
@kris-lab nope, should not (see https://github.com/cargomedia/puppet-packages/pull/832/files#diff-0f87f2a21a1b2a811cc7e7ea5d125749R22)

njam:
I think this is a very good change!

- How about renaming `helper::service` to `sysvinit::script`?
 - Further down the road we could even have a default *template* for the script, and rather pass parameters like `$executable`, `$pidfile` etc...
- helper/manifests/service.pp: I would keep the `unless` (and drop `refreshonly`). Defining the *state* we want some resources to be in is one of the big advantages that makes puppet code robust. Even if there are other processes fiddling with these resources, we can make sure that we will get to the desired state. How about using [`invoke-rc.d --query`](https://github.com/puppetlabs/puppet/blob/cb4ad1f19b043a70ba17e4103a25f5c97a76948b/lib/puppet/provider/service/debian.rb#L34)?
- Monit init script: You said monit's init script doesn't need to be installed with `update-rc.d`, it would start because of `/etc/monit/default`? I understand this default-file is read in the init-script to enable or disable actual startup. But the init-script itself still needs to be installed. In this case probably from the debian package?

ppp0:
* renaming.. always difficult, the naming.. what about: `init::sysvinit::script` ?
* I don't completely get how to reliably use `invoke-rc.d`. The man page states that it returns codes 100 - 105 with this meaning:

>100    Init script ID (name) unknown.  This means the init script was not registered successfully through update-rc.d or that  the  init  script  does not  exist.
101     Action not allowed.  The requested action will not be performed because of runlevel or local policy constraints.
102     Subsystem error.  Init script (or policy layer) subsystem malfunction. Also, forced init script execution due to --try-anyway or --force failed.
103    Syntax error.
104    Action allowed.  Init script would be run, but --query is in effect.
105    Behavior uncertain.  It cannot be determined if action should be carried out or not, and --query is in effect.
106    Fallback action requested.  The policy layer denied the requested action, and supplied an allowed fallback action to be used instead.

So let's try it in practice:
```
root@www:/home/vagrant# invoke-rc.d memcached status;echo $?
[ ok ] memcached is running.
0
root@www:/home/vagrant# invoke-rc.d --query memcached status;echo $?
105
```

`105` being `behaviour uncertain.... ?

this one is installed for our script:

```
root@www:/home/vagrant# invoke-rc.d --query nginx stop;echo $?
105
root@www:/home/vagrant# invoke-rc.d --query nginx start;echo $?
104
root@www:/home/vagrant# invoke-rc.d --query nginx status;echo $?
105
```

Tbh, I am confused... What's the problem in calling `update-rc.d` whenever the init file is changed? Don't we get to a "desired state"?

* `monit`: I must confess I haven't dug too deep into the matter. So using the helper script would be ok to me for monit, too

njam:
> * renaming.. always difficult, the naming.. what about: `init::sysvinit::script` ?

So you propose a module `init` - which would contain what apart `sysvinit`? Don't we usually have 1 module per software (which in this case is `sysvinit`)?

> I don't completely get how to reliably use `invoke-rc.d`

Agree, as can be seen [in puppetlabs' code](https://github.com/puppetlabs/puppet/blob/cb4ad1f19b043a70ba17e4103a25f5c97a76948b/lib/puppet/provider/service/debian.rb#L34) they also fall back to checking `/etc/rc*.d` in case of ext code 105:
```rb
Dir.glob("/etc/rc*.d/S??#{@resource[:name]}").length
```
Let's stick to file check then.

> What's the problem in calling `update-rc.d` whenever the init file is changed? Don't we get to a "desired state"?

Problem is we don't really *manage* that resource (init script installation) anymore then, because we never check it's state. So any other component (in puppet, or from other software) can change this resource, and we will not notice and not do anything about it.
With puppet we have this ability to define exactly in which state a resource should be, and puppet will make sure it is like that every 10 minutes, which brings great robustness, in the sense that we can trust all resources are in the desired state if the puppet run succeeds?

ppp0:
* Rationale is that there could be a `init::upstart` `init::systemd`, etc etc
* I understand.. Good to go with globbing on `rc*.d/`.. I like

njam:
> * Rationale is that there could be a `init::upstart` `init::systemd`, etc etc

I see. I would rather keep the two apart. While they solve similar problems, it's two different pieces of software. Putting them into 1 module will make things complicated I think (like having `apache2` and `nginx` in 1 module?).

tomaszdurka:
I would gladly review, pair, etc this. @ppp0 let me know if you want to pair on it?

ppp0:
@tomaszdurka please review

modules/sysvinit/manifests/script.pp
```diff
@@ -0,0 +1,25 @@
+define sysvinit::script(
+  $init_file_content,
```
tomaszdurka: Call this content?

modules/sysvinit/manifests/script.pp
```diff
@@ -0,0 +1,25 @@
...
+  ~>
+
+  exec { "update-rc.d ${name} defaults":
+    path        => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
+    refreshonly => true,
+  }
+  ~>
+
+  exec { "/etc/init.d/${name} start":
+    path        => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
```
tomaszdurka: Do you need paths if exec path is absolute?

modules/sysvinit/manifests/script.pp
tomaszdurka: Another question, should we have it so open that you can set any content or we should rather allow to pass specific variables like:
DAEMON, DAEMON_ARGS, USER, etc...?

ppp0: Yes, it will call this content :stuck_out_tongue:

modules/sysvinit/manifests/script.pp
```diff
@@ -0,0 +1,25 @@
...
+  }
+  ~>
+
+  exec { "update-rc.d ${name} defaults":
+    path        => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
+    refreshonly => true,
+  }
+  ~>
+
+  exec { "/etc/init.d/${name} start":
```
tomaszdurka: Should we make this configurable? `if $autostart`

ppp0: See https://github.com/cargomedia/puppet-packages/pull/832#issuecomment-70878976

modules/sysvinit/manifests/script.pp
ppp0: agree, iirc we had a discussion about how to format `exec { }` and that we always use the same `path` array?, where do you see an absolute path? `init.d/${name}` ? Tbh I don't know if the script wants to have a `$PATH` variable set in the environment?

modules/sysvinit/manifests/script.pp
tomaszdurka: Ok, btw aren't we reinventing wheel.

When I look at service docu https://docs.puppetlabs.com/references/latest/type.html#service seems it has similar capabilities (`$enable`). I know we already use `service` to restart services.

modules/sysvinit/manifests/script.pp
tomaszdurka: Hmm? :panda_face:

ppp0:
@tomaszdurka any need to pair on this still?

Otherwise please review!

Note that I've refactored all occurences of `Service` to `ensure => true`

modules/sysvinit/manifests/script.pp
```diff
@@ -1,14 +1,20 @@
...
 
-  exec { "update-rc.d ${name} defaults":
-    path   => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
-    unless => "ls /etc/rc3.d/ | grep -- '^S..${name}$'",
+  file { "/etc/init.d/${name}":
+  ensure  => file,
+  content => $content,
+  owner   => '0',
+  group   => '0',
+  mode    => '0755',
```
tomaszdurka: Formatting

modules/pulsar_rest_api/manifests/init.pp
```diff
@@ -86,7 +86,8 @@
     refreshonly => true,
   }
 
-  helper::service { 'pulsar-rest-api':
+  sysvinit::script { 'pulsar-rest-api':
+    content => template("${module_name}/init.sh"),
```
tomaszdurka: Remove https://github.com/ppp0/puppet-packages/blob/issue-832/modules/pulsar_rest_api/manifests/init.pp#L62

modules/mms/manifests/agent/monitoring.pp
```diff
@@ -16,35 +16,29 @@
...
+    mode    => '0644',
+    require => Helper::Script['install-mms-monitoring'],
+    notify  => Service[$agent_name];
   }
   ->
 
-  helper::service{ $agent_name:
+  sysvinit::script { $agent_name:
+    content           => template("${module_name}/init"),
+    subscribe         => File["/etc/init.d/${agent_name}"],
```
tomaszdurka: I think this is not needed too? (will be created internally - probably historical)

modules/copperegg_revealcloud/manifests/init.pp
```diff
@@ -67,20 +58,20 @@
       refreshonly => true,
       user        => '0',
       group       => '0',
-      require     => File['/etc/init.d/revealcloud'],
+      require     => Sysvinit::Script['revealcloud'],
       before      => Service['revealcloud'],
     }
 
-    Exec['update-rc.d revealcloud defaults'] ~> Exec['enable revealcloud node']
+    Sysvinit::Script['revealcloud'] ~> Exec['enable revealcloud node']
```
tomaszdurka: Maybe simply add `subscribe` for `exec { 'enable revealcloud node' }` instead of this line?

modules/sysvinit/manifests/script.pp
```diff
@@ -0,0 +1,23 @@
...
+    owner   => '0',
+    group   => '0',
+    mode    => '0755',
+  }
+  ~>
+
+  exec { "/etc/init.d/${name} start":
+    path        => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
+    unless      => "/etc/init.d/${name} status",
+    refreshonly => true,
```
tomaszdurka: This should also notify service? Maybe add this at line 21-22?

modules/cgroups/manifests/init.pp
```diff
@@ -28,19 +28,10 @@
...
-    notify  => Service['cgconfig-apply'],
-  }
-  ~>
-
-  exec { 'update-rc.d cgconfig-apply defaults && /etc/init.d/cgconfig-apply start':
-    path        => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
-    refreshonly => true,
+  sysvinit::script { 'cgconfig-apply':
+    content           => template("${module_name}/init"),
+    notify            => Service['cgconfig-apply'],
```
tomaszdurka: Shouldn't this be done inside `sysvinit::script`?

ppp0:
@tomaszdurka one last review, please

tomaszdurka:
`pulsar_rest_api` `Puppetfile` missing sysvinit dependency.

Apart from that looks good.

ppp0:
@tomaszdurka

tomaszdurka:
Looks good, but let's wait for the build! :panda_face:

njam:
retest this please

ppp0:
retest this please

njam:
retest this please

ppp0:
retest this please

ppp0:
#hugesuccess
@tomaszdurka please give you OK

tomaszdurka:
Looks good then!

PR #836, opened 2015-01-27
ruby::gem::nokogiri fail to install

kris-lab:
`nokogiri` on debian requires the lib `zlib1g-dev` [see](http://www.nokogiri.org/tutorials/installing_nokogiri.html)

### DIFF
--- a/modules/ruby/manifests/gem/nokogiri.pp
+++ b/modules/ruby/manifests/gem/nokogiri.pp
@@ -2,7 +2,7 @@ class ruby::gem::nokogiri ($version = 'present') {
 
   require 'build'
 
-  package {['libxslt1-dev', 'libxml2-dev']:
+  package {['libxslt1-dev', 'libxml2-dev', 'zlib1g-dev']:
     ensure => present
   }
 


kris-lab:
@njam @ppp0 please have a look; this gem fails on our machines without added lib

ppp0:
Uh oh... is that all libs that are required or are there still some which we assume but don't necessarily need to be there?

What about the env variable `export NOKOGIRI_USE_SYSTEM_LIBRARIES=1` ?

njam:
> NOKOGIRI_USE_SYSTEM_LIBRARIES

They recommend not using this option, because some library versions might bring problems. I'd stick with their recommendation.

---

I can install nokogiri locally without `zlib1g-dev`. Why is it needed? When does it fail?

ppp0:
May be on SoftLayer they don't include it in the minimal install?

njam:
Makes sense!

kris-lab:
It failed on 2 new machines today. I did install it manually to finish provisioning!

njam:
Agree, good to go imo!

ppp0:
Oh, for me too! LGTM

PR #840, opened 2015-01-28
Apply formatting

njam:
Paired with @tomaszdurka


njam:
Build killed on purpose.

PR #841, opened 2015-01-28
Add bipbip entry for socket-redis

kris-lab:
Depends on https://github.com/cargomedia/bipbip/pull/100

### DIFF
--- a/modules/socket_redis/manifests/init.pp
+++ b/modules/socket_redis/manifests/init.pp
@@ -127,4 +127,10 @@ class socket_redis (
     content => template("${module_name}/monit"),
     require => Service['socket-redis'],
   }
+
+  @bipbip::entry { 'socket-redis':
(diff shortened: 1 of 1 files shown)


kris-lab:
@njam please review

njam:
- Dash is not allowed in resource name?
- Use host+port config and pass it on to bipbip

kris-lab:
* ok
* ok: socket-redis host is not a class arguments so will use localhost (only redis host is known)

kris-lab:
@njam please review

njam:
- Now you can drop the `plugin` key
- url key should be prefixed with `http://`?

kris-lab:
* we use `plugin` key everywhere!?
* `http://` is not needed imo; it uses http transportation by default

kris-lab:
@njam please review?

njam:
lgtm!

PR #843, opened 2015-01-30
Upgrade s3export_backup

tomaszdurka:


ppp0:
lvgtm

tomaszdurka:
As agreed these tests will fail until serverspec upgraded.

njam:
What's the problem? Bug in the serverspec we use? Add a note to the test?

tomaszdurka:
Let's wait for serverspec upgrade - it should solve it.

PR #847, opened 2015-02-05
Add bower module

njam:


njam:
Paired with @tomaszdurka

PR #851, opened 2015-02-17
Introduce configurable redirects for cm::vhost

ppp0:


ppp0:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/613

@njam please review.. 

I'm not sure if a redirect should listen on port `443`, too - it probably should?

njam:
- Should work opposite: There should be a list of `redirects`, which is a list domains that will redirect back to the original domain.
- Redirect should be for port 80 only (it's merely about typos, and we don't want to buy certs)

ppp0:
* ooops
* ok

@njam please review

njam:
- The https-redirect should also be applied for the original domain if $redirect is present? (add a spec?)

ppp0:
makes sense makes sense
please review @njam

njam:
Looks good!

PR #857, opened 2015-02-20
Add HSTS to cm::vhost

kris-lab:
Some notes:
- https://coderrr.wordpress.com/2010/12/27/canonical-redirect-pitfalls-with-http-strict-transport-security-and-some-solutions/
- http://linux-audit.com/configure-hsts-http-strict-transport-security-apache-nginx/

### DIFF
--- a/modules/cm/manifests/vhost.pp
+++ b/modules/cm/manifests/vhost.pp
@@ -25,6 +25,9 @@ define cm::vhost(
       location_cfg_append => [
         "return 301 ${protocol}://${name}\$request_uri;",
       ],
+      vhost_cfg_prepend   => [
+        'add_header Strict-Transport-Security "max-age=31536000; includeSubdomains;";'
+      ]
     }
   }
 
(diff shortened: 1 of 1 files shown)


kris-lab:
@njam have a look;

tested on staging; seems to be working good!

njam:
Should this only be enabled when using SSL?
What effect does the "max-age" have?

kris-lab:
- agree
- "Max-age limits should be carefully considered as infrequent visitors may find your site inaccessible if you relax your policy." so in my understanding if we change back to use HTTP then browser could cache it and make inaccessible? ([see](https://www.owasp.org/index.php/HTTP_Strict_Transport_Security))

@njam please re-review; (wdyt about passing empty string if ssl not enabled?)

njam:
- So maybe reduce "max-age" a bit? 10 days?
- What about `includeSubdomains`? Since we're setting this for `example.com` this will also affect other sub domains that we possibly *don't* manage with this vhost (e.g. `ads.example.com` etc)
- I can see that our AWS CloudFront CDN is pulling over `http://` from the "cdn_origin_vhost".

Overall I now too have the feeling this is a bit dangerous due to the above if we don't thoroughly test it. Should we skip it?

kris-lab:
Agree. Let's skip for now. 

closing.

PR #859, opened 2015-02-20
Introduce hash-based argument-passing for cm::vhost

ppp0:
As a follow-up of #851 

We want to be able to pass a `redirects` list to arbitrary `hostnames`..

Example (hiera):
```
    "sk::www::vhosts": [
      {
        "domain": "foo.com",
        "aliases": ["www.foo.com", "admin.foo.com"]
      },
      {
        "domain": "oh.net",
        "aliases": ["www.oh.net", "admin.oh.net"],
        "redirects": ["ah.net", "uh.net"],
      }
    ]
```

### DIFF
--- a/modules/cm/manifests/vhost.pp
+++ b/modules/cm/manifests/vhost.pp
@@ -3,7 +3,7 @@ define cm::vhost(
   $ssl_cert = undef,
   $ssl_key = undef,
   $aliases = [],
-  $redirect = undef,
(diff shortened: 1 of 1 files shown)


njam:
I tried a bit and it seems one cannot pass array of hashes to a custom function.
Reverting the original change for now: https://github.com/cargomedia/puppet-cargomedia/pull/744

PR #860, opened 2015-02-26
Configure puppet-master / librarian-puppet to rsync by default

kris-lab:
part of https://github.com/cargomedia/puppet-cargomedia/issues/746

### DIFF
--- a/modules/librarian_puppet/manifests/config.pp
+++ b/modules/librarian_puppet/manifests/config.pp
@@ -16,7 +16,7 @@ define librarian_puppet::config (
   $command = "${command_prefix} ${value} ${command_option}"
   $unless = "${command_prefix} | grep ${name}"
 
-  exec { "set config ${name} to ${value}":
+  exec { "set librairan-puppet config ${name} to ${value}":
     command     => $command,
     path        => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
     provider    => shell,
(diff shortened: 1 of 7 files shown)


kris-lab:
@njam wdyt? maybe it should go to librarian module and be configured globally for root?

kris-lab:
retest this please

njam:
- Shouldn't this go to `puppet::master::puppetfile`?
- Spec: Check whether config is really set and applied with `librarian-puppet config KEY`?

kris-lab:
@njam what bout introducing `librarian_puppet::config`? The code should be tuned and optimised but wdyt in general?

njam:
lvgtm, I would:
- replace $environment with $user_home (because it's the only usage?)

kris-lab:
@njam please review

njam:
- Could we use `exec`'s `cwd` argument instead of `cd`'ing?
- specs fail

kris-lab:
@njam please review; only puppetdb test fails!

njam:
One more thing (sorry):
- puppet::master::puppetfile: I'm not sure about the `$librarian_config` param, it requires to know internals of the class (the `/etc/puppet` path). How about just hardcoding these settings in the class? I don't think any other settings make sense?

kris-lab:
@njam please review

njam:
- `librarian_puppet::config`'s title should be `rsync for /etc/puppet` to avoid conflicts?

Rest look sgood!

kris-lab:
agree. @njam please re-review...

njam:
Looks good!
Let's hope this helps!

PR #864, opened 2015-03-03
Remove string to hash/array conversion

tomaszdurka:
Expect stringify_facts=false

https://tickets.puppetlabs.com/browse/PUP-2966

- set config for puppet
- network interfaces (optional)


tomaszdurka:
Paired with @kris-lab

PR #865, opened 2015-03-03
Rework copperegg tags to work with facts

tomaszdurka:


tomaszdurka:
Paired with @kris-lab

PR #835, opened 2015-03-09
Make "monit-alert.sh" wait for (optional) monit restart

tomaszdurka:
In case of problems detecting monit's `reload` in `monit-alert.sh` we do a `restart` instead:
```sh
if ! (timeout --signal=9 5 bash -c "while ! (monitCheckHasReloaded); do sleep 0.05; done"); then
  # Hard-restart monit in case it cannot reload
  /etc/init.d/monit restart
fi
```

This restart is not blocking, so the next script to run might find a stopped monit. We should make sure monit is fully restarted after the script finished.

### DIFF
--- a/modules/monit/templates/bin/monit-alert.sh
+++ b/modules/monit/templates/bin/monit-alert.sh
@@ -17,16 +17,16 @@ ln -s ${SOURCE} ${DEST}
 function monitGetStateFileAccess { stat -c "%Y" "/root/.monit.state"; }
 export -f monitGetStateFileAccess
 export MONIT_STATE_FILE_ACCESS=$(monitGetStateFileAccess)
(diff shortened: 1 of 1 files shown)


tomaszdurka:
I've pushed all the code - it still needs verification though.

ppp0:
Not sure if:
* 10 seconds after restart a bit too large?
* check for state change after a restart inside the first `if..fi` block instead of repeating the check in `30..32`?

tomaszdurka:
- well if it's quicker it will finish quicker
- done

ppp0:
@tomaszdurka agree to the changes aka please review

I tested intensively on VM with all possible failures

tomaszdurka:
Looks good.

PR #868, opened 2015-03-10
Allow to pass tags for bipbip

njam:
@kris-lab please review


kris-lab:
what happens if `$::copperegg_tags` fact is not set? is it undef/nil?

njam:
yes, undef, will be handled in template!

kris-lab:
so lgtm

Issue #837, opened 2015-01-27
tests failing for ruby::gem::bipbip


ppp0:

ppp0:
Failure was:
```
Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install libmysqlclient-dev' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
The following extra packages will be installed:
  libmysqlclient18 mysql-common
The following NEW packages will be installed:
  libmysqlclient-dev libmysqlclient18 mysql-common
0 upgraded, 3 newly installed, 0 to remove and 29 not upgraded.
Need to get 1694 kB of archives.
After this operation, 9337 kB of additional disk space will be used.
Err http://security.debian.org/ wheezy/updates/main mysql-common all 5.5.40-0+wheezy1
  404  Not Found [IP: 127.0.0.1 8123]
Err http://security.debian.org/ wheezy/updates/main libmysqlclient18 amd64 5.5.40-0+wheezy1
  404  Not Found [IP: 127.0.0.1 8123]
Err http://security.debian.org/ wheezy/updates/main libmysqlclient-dev amd64 5.5.40-0+wheezy1
  404  Not Found [IP: 127.0.0.1 8123]
Failed to fetch http://security.debian.org/pool/updates/main/m/mysql-5.5/mysql-common_5.5.40-0+wheezy1_all.deb  404  Not Found [IP: 127.0.0.1 8123]
Failed to fetch http://security.debian.org/pool/updates/main/m/mysql-5.5/libmysqlclient18_5.5.40-0+wheezy1_amd64.deb  404  Not Found [IP: 127.0.0.1 8123]
Failed to fetch http://security.debian.org/pool/updates/main/m/mysql-5.5/libmysqlclient-dev_5.5.40-0+wheezy1_amd64.deb  404  Not Found [IP: 127.0.0.1 8123]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
[...]
```

`apt-get update` was needed.

This is no bug, as nothing can be done about it (except for requiring `apt-get update` to run on every `Package` resource inclusion?

I had to 
`vagrant destroy -f` to get rid of all snapshots and then `apt-get update` was invoked again as a dependency of https://github.com/cargomedia/puppet-packages/blob/master/spec/provision.pp#L4

njam:
[...]

PR #838, opened 2015-03-10
Run apt-get update if local package cache older than x

kris-lab:
Following the discussion in #837 

To be confirmed if it is a viable solution

e.g. (ruby)
```rb
 (Time.now - File.stat('/var/cache/apt').mtime).to_i / 86400.0
```
returns the age of the directory in days as `float` 

refresh every day?


njam:
Also now we could require `apt::update` when installing any package with apt?
Maybe also check usages of `apt::update`, I remember we did some quirky things there to make sure the index is up to date in the right moment..

kris-lab:
afaics we require `apt update` with `exec` for each `package` now?

njam:
Makes sense, as soon as we either include `apt::update` or `apt`. Which we do in `cargomedia::base`.
But we don't do it in puppet-packages tests?

Looking at the code I would
- Use seconds for age (SI unit), or make it clear it's days

kris-lab:
@ppp0 @njam could you have a look on spec?

The problem is that `apt::update` running command `apt-get -o Acquire::http::proxy=false update` is not updating modification timestamp for `/var/cache/apt`. If I run it manually in the box then all is fine. If puppet runs this command then it is done but time of access/modification is not updated?

njam:
Try this approach?
http://serverfault.com/questions/20747/find-last-time-update-was-performed-with-apt-get/425237#425237

kris-lab:
Still the same problem!

kris-lab:
@njam please review

njam:
renaming as discussed, rest looks good

ppp0:
very good job!

PR #870, opened 2015-03-11
Improve monit-alert script

njam:
- Send SIGHUP to process
- Check syslog to detect reload

Paired with @kris-lab


njam:
@tomaszdurka @ppp0 fyi

PR #872, opened 2015-03-12
Fix encoding for pulsar-rest-api/pulsar

tomaszdurka:


tomaszdurka:
Paired with @kris-lab

PR #876, opened 2015-03-19
Increase monit's default load alerting thresholds

njam:
See e.g. http://blog.scoutapp.com/articles/2009/07/31/understanding-load-averages
Related to https://github.com/cargomedia/puppet-cargomedia/issues/659

@cargomedia/devops wdyt?

### DIFF
--- a/modules/monit/manifests/entry/system.pp
+++ b/modules/monit/manifests/entry/system.pp
@@ -1,6 +1,7 @@
 class monit::entry::system(
   $alertOnLoad1 = undef,
-  $alertOnLoad5 = undef
+  $alertOnLoad5 = undef,
+  $alertOnLoad15 = undef
 ) {
 
   require '::monit'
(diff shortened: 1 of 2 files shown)


ppp0:
if it helps, it's good - the topic is complex!

njam:
So to summarize: a load of 1 means that the CPU is working at full capacity all the time, but no processes are waiting for execution time. Of course each processes' need of CPU calculations fluctuates a lot, so it can happen that many processes wait for CPU execution (number of processes in the run queue).

I've now reduced the 15 mins threshold to 0.8.
But for 1 minute we can go up to 5, and for 5 minues up to 2.

@cargomedia/devops wdyt?

ppp0:
It's ok for me

tomaszdurka:
Ok with me (reading up your linked resources).

kris-lab:
sgtm

PR #880, opened 2015-04-07
Pin puppetdb version until migration bug fixed

ppp0:


ppp0:
@njam please review - I think this is less cumbersome as a clean install.. for now!

njam:
lgtm.

What's the plan for resolution?

ppp0:
1) As discussed, shall we stick with puppetdb, then do a clean install in a slack moment (needs several puppet agent runs on nodes and supervision that all is ok - not sure if it is straightforward to do, though)
2) Upgrade to postgres (https://github.com/cargomedia/puppet-cargomedia/issues/780)

PR #882, opened 2015-04-13
Allow to use SSL for cm::services

njam:


njam:
@ppp0 please review

ppp0:
How about grouping all packages to be spec tested in a list. E.g.

```ruby
packages=['redis-server','mysql-server','memcached']
packages.each do |pkg|
    describe package(pkg) do
     it { should be_installed }
   end
end
```

to make it more concise (and readable)?

njam:
I agree in general, but not for this specific use case :)

I think these checks for installed packages are temporary and should be replaced by different assertions. Because in `cm::services` we shouldn't check for specific packages, but rather that some services are running, ports are open etc.
I didn't do it now, because it wasn't the focus of my PR.
What do you think? How would you test the `cm::services`?

ppp0:
hm, understand.. what about having a client testing that services are answering (e.g. `curl` would do for most cases?)

njam:
Maybe there should be a command in CM: `bin/cm app check-environment`.

Anyway I think that's a different topic. I would suggest to keep it as it is now, wdyt?

ppp0:
lgtm

PR #892, opened 2015-04-30
Disable bipbip log rotation temporarily

njam:
To investigate https://github.com/cargomedia/bipbip/issues/114
@tomaszdurka please review


tomaszdurka:
Looks ok.

njam:
Applied to production  with:
```sh
sudo rm /etc/logrotate.d/bipbip
```

PR #893, opened 2015-04-30
Run "apt update" before every spec

njam:
..to make sure apt's index is up-to-date.

This adds 2.5secs for every test.

@ppp0 @tomaszdurka wdyt?

### DIFF
--- a/spec/lib/puppet_spec.rb
+++ b/spec/lib/puppet_spec.rb
@@ -47,36 +47,26 @@ class PuppetSpec
     @spec_dir.entries.sort.each do |relative_path|
       next unless relative_path.extname == '.pp'
 
-      manifest_path = relative_path.expand_path(@spec_dir)
-      manifest_path_vagrant = @vagrant_box.parse_external_path(manifest_path)
-      run_apply_manifest(manifest_path_vagrant)
+      run_apply(@vagrant_box.working_dir.join('spec/spec_before.pp'))
+      run_apply(relative_path.expand_path(@spec_dir))
     end
   end
 
   private
 
-  # @param [Pathname] path
-  def run_apply_manifest(path)
-    run_apply(path.to_s.shellescape)
-  end
-
-  # @param [String] puppet_code
-  def run_apply_code(puppet_code)
-    run_apply("--execute #{puppet_code.shellescape}")
-  end
-
-  # @param [String] arg
-  def run_apply(arg)
+  # @param [Pathname] manifest_path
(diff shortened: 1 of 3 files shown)


ppp0:
I reckon there is no other way.. sounds a bit redundant - placing the apt cache on a shared folder (`/vagrant/spec/apt/`)could be an option?

Looking at the code changes I'm not sure it this issue's title completely reflects what you are doing

njam:
Yup, I've also changed the puppet run to use `--detailed-exitcodes`.

I'd suggest to merge this version for now to avoid failing builds, and look into the shared folder option later as an optimization.
wdyt?

ppp0:
sgtm, lgtm

njam:
@tomaszdurka wanna have a look too?

tomaszdurka:
Are you sure that `--detailed-exitcodes` will make puppet command fail on puppet resources failures? I remember that regexp search was added on purpose. If catalog was compiled fine command always returned 0 no matter if there were some puppet errors during apply.

njam:
I had something like that in mind too, but couldn't confirm.
Now I think I remember: If the catalog compilation fails it didn't error? I will test this...

tomaszdurka:
If remember correctly puppet apply exit code is related only to compilation of the catalog, but not with realization of the compiled resources.

- It will return non-zero code on syntax errors, infinite dependency loops, etc.
- It will return zero code on resource realization errors.

njam:
Can't confirm (anymore): Errors during compilation and during applying both lead to non-0 exit code.
```
# puppet apply -e 'package {"dshfihn":}' --detailed-exitcodes; echo $?
Notice: Compiled catalog for debian-7-amd64-default.vagrantup.com in environment production in 0.11 seconds
Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install dshfihn' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package dshfihn
Error: /Stage[main]/Main/Package[dshfihn]/ensure: change from purged to present failed: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install dshfihn' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package dshfihn
Notice: Finished catalog run in 0.25 seconds
4
```

Agree to merge then?

tomaszdurka:
Sure. I would keep a closer look on specs for some time.

PR #896, opened 2015-05-05
Increase monit reload timeout from 10 to 30 secs

njam:
When monit is busy with starting/stopping daemons (e.g. after starting up), it can take a bit to process the SIGHUP and reload

This can be reproduced for example with:
```
/etc/init.d/monit stop
monit stop puppet
/etc/init.d/monit start
monit-alert default
```


njam:
@ppp0 @kris-lab please review

kris-lab:
lgtm

PR #897, opened 2015-05-08
Add ssh-agent start in pulsar-rest-api init script

njam:


njam:
@tomaszdurka @kris-lab please review

You can run `ssh-agent <foo>`, then the agent will run the command `foo` in an environment with all the variables set up. Plus this approach makes sure the agent only runs as long as the daemon is running (basically the agent *is* the daemon now).

@vogdb fyi

kris-lab:
lvgtm

PR #908, opened 2015-06-26
jenkins::plugin doesn't check installed plugin's version

njam:
Possibly use something like:
```
cat /var/lib/jenkins/plugins/<PLUGIN>/META-INF/MANIFEST.MF | grep 'Plugin-Version:'
```


njam:
@ppp0 please review

ppp0:
lgtm

PR #917, opened 2015-07-20
Fix lvm spec

ppp0:
Secondary disk in virtualbox (`/dev/sdb`) was removed here: https://github.com/cargomedia/puppet-packages/pull/916
(The problem with the previous approach was that the `Vagrantfile` is declarative, so we can't check if the disk's file already exists with ruby, and then call `vb.customize ['createhd' ...` if not, because it will run this customization multiple times.)

@ppp0 maybe there's a way to refactor the LVM spec to work differently? Could we create a "pseudo disk" device on the fly?

### DIFF
--- a/modules/lvm/spec/default/manifests.pp
+++ b/modules/lvm/spec/default/manifests.pp
@@ -1,7 +1,18 @@
 node default {
 
+  package { ['mount', 'util-linux']:
+    ensure => present,
+  }
+  ->
+
+  helper::script { 'Setup temporary loop device':
+    content => 'TMP=$(mktemp);dd if=/dev/zero of=${TMP} bs=100 count=1M;losetup /dev/loop0 ${TMP}',
+    unless  => "lsblk | grep -q loop0",
+  }
+  ->
+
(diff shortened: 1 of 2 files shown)


ppp0:
Now using a loop disk out of a temp file. Min. size must be around 40MB for xfs filesystem. 10MB did not work. Using 100MB, did not fully investigate

@njam please review

njam:
Syntax check is failing.
Rest looks good.

ppp0:
@njam please review

ppp0:
These checks were failing:
```
modules/lvm/spec/default/manifests.pp - WARNING: double quoted string containing no variables on line 10
modules/lvm/spec/default/manifests.pp - ERROR: single quoted string containing a variable found on line 9
```

njam:
The first line has *bash* variables I guess, so needs to use single quotes *or* escaping?

ppp0:
Changes bash variable expansion to alternative syntax

@njam please review

njam:
lgtm

PR #919, opened 2015-07-27
Allow configuration of relay SMTP servers

ppp0:
Right now, we are able to relay to specific SMTP servers based on a pattern-matching mechanism applied to the mails headers. In some cases, a general (aka catch-all) mechanism is preferred.

As of now we can do: 
```puppet
  class { 'postfix':

    aliases    => {
      'root' => 'another@mail.com',
      'fuboo' => 'root'
    },

    transports => [
      {
        'filter' => '/^X-MDA: Amazon$/',
        'protocol' => 'smtp',
        'host' => 'mail.amazon.com',
        'port' => 'submission',
        'credentials' => 'foo:bar',
      }
    ]
  }
```

The `transports` list would create `/etc/postfix/header_checks` and `/etc/postfix/sasl_passwd` to correctly route mails based on the filter matched.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,16 +3,17 @@ GEM
   specs:
     CFPropertyList (2.2.8)
     diff-lcs (1.2.5)
-    facter (2.4.1)
+    facter (2.4.4)
       CFPropertyList (~> 2.2.6)
     hiera (1.3.4)
       json_pure
     json_pure (1.8.2)
-    multi_json (1.10.1)
+    multi_json (1.11.2)
     net-scp (1.2.1)
       net-ssh (>= 2.6.5)
     net-ssh (2.9.2)
-    puppet (3.7.4)
+    net-telnet (0.1.1)
+    puppet (3.8.1)
       facter (> 1.6, < 3)
       hiera (~> 1.0)
       json_pure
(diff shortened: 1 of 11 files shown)


ppp0:
@cargomedia/devops there are so many implementation possibilities for this ["catch-all"](https://github.com/cargomedia/puppet-cargomedia/issues/837#issuecomment-122820096)

The easiest really is to use a pattern that will always match when regex'ed on the headers.. (such as `//`) and to filter that to a relay smtp server.

Another possibility would be to leave `{ transports['filter'] }` empty and replace it said `//` in the template

See spec `test:postfix:relay`

I tested on VM with 100 custom headers and could not measure a significant performance impact

wdyt?

@njam please review

njam:
Tests! :confetti_ball:

- relay/manifest.pp::30: Formatting

Rest looks good

ppp0:
Good, so now we can do this:

```pp
[..]
    transports => [
      {
        'protocol' => 'smtp',
        'host' => 'mail.amazon.com',
        'port' => 'submission',
        'credentials' => 'foo:bar',
      }
    ]
[..]
```

I.e. leave the `filter` property alone, then a catch-all will be implemented.

**Caveat:** If several `transports` are defined, postfix will not stop after one rule has matched but instead continue to try to match. This means: of a chain of matching patterns, the last one will finally be authoritative!

ppp0:
@njam oh oh, I inadvertently pushed the updated `Gemfile` too. IMO this is ok, dytst? (do you think so too? :) )

njam:
Sure, let's see what jenkins has to say about it!

ppp0:
looks good

PR #924, opened 2015-07-29
autoprefixer: Make sure we upgrade caniuse-db automatically

njam:


njam:
@kris-lab please review

We install the npm package `autoprefixer` which has a dependency to `caniuse-db`. The latter should *always* be updated to the latest version.
Since `caniuse-db` is not the global system package installed, but a local one within `autoprefixer` we can't used the global `npm` package provider. I'm introducing a new define `nodejs::package` which accepts a path where to install the package, a name and a version. It also allows to install `latest`.
(it's inspired by this code: https://github.com/puppet-community/puppet-nodejs/blob/master/manifests/npm.pp)

@christopheschwyzer @vogdb fyi

njam:
@vogdb or @christopheschwyzer could you check that the site works good with autoprefixer `5.2.0` and caniuse-db `1.0.30000247`?

vogdb:
I have problems on Android 4.2 but that is possible due to the error `Undefined class constant 'RELATIVE_PATHS'` which I still have >_<. I can't even complete my example setup script `./sk/bin/cm app setup --reload`. But iOS and Chrome work well.

kris-lab:
@njam good idea. code looks good!

christopheschwyzer:
What's the easiest way to test this?

vogdb:
@christopheschwyzer  There is no one, I guess. We need to have a list of places where we depend on the correct prefixing.

vogdb:
Nope, it's not the problem of autoprefixer. It's a common Android problem. The same on production. So, here I see lgtm. The problem I found has a different root and is posted here cargomedia/sk#1916.

PR #925, opened 2015-08-06
Add logparser to watch mysql error log for crashed tables

ppp0:
First proposed in https://github.com/cargomedia/puppet-cargomedia/pull/852

### DIFF
--- a/modules/mysql/manifests/server.pp
+++ b/modules/mysql/manifests/server.pp
@@ -123,4 +123,16 @@ class mysql::server ($root_password = '', $debian_sys_maint_password = '') {
       'password' => $debian_sys_maint_password,
     }
   }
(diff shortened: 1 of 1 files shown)


ppp0:
@kris-lab can you please review? 

From what I understand, in https://github.com/cargomedia/copperegg-revealmetrics/blob/master/lib/copperegg/revealmetrics/metric_group.rb#L71 the correct HTTP method (put) is used, so adding an additional metric to an existing metric group (`mysql` in this case) should work

kris-lab:
afaics lvgtm. metrics group should be extended automatically. dashboard must be updated manually!?

ppp0:
Updated dashboard. Looks good!

PR #926, opened 2015-08-10
logparser failed access to mysql error log

ppp0:
introduced in #925 

Error:
```
E, [2015-08-06T15:54:35.353479 #22925] ERROR -- log-parser ads2.fuckbook.com::mysql::_var_log_mysql_error_log: Permission denied - /var/log/mysql/error.log
```

Access rights to `/var/log/mysql` are wrong:

```
root@ads2.fuckbook.com:/home/admin$ ls -al /var/log/mysql
total 36
drwxr-s---  2 mysql adm  4096 Aug  6 06:25 .
drwxr-xr-x 16 root  root 4096 Aug  6 06:25 ..
-rw-r-----  1 mysql adm     0 Aug  6 06:25 error.log
```

### DIFF
--- a/modules/mysql/manifests/server.pp
+++ b/modules/mysql/manifests/server.pp
@@ -116,7 +116,7 @@ class mysql::server ($root_password = '', $debian_sys_maint_password = '') {
     ]
   }
 
-  logrotate::entry{ 'mysql-server-error':
+  logrotate::entry { 'mysql-server-error':
     content => template("${module_name}/logrotate-error"),
     before  => Package['mysql-server'],
   }
--- a/modules/mysql/manifests/server.pp
+++ b/modules/mysql/manifests/server.pp
@@ -116,6 +116,11 @@ class mysql::server ($root_password = '', $debian_sys_maint_password = '') {
     ]
   }
 
+  logrotate::entry{ 'mysql-server-error':
+    content => template("${module_name}/logrotate-error"),
+    before  => Package['mysql-server'],
+  }
+
   @monit::entry { 'mysql':
     content => template("${module_name}/monit"),
     require => Service['mysql'],
--- a/modules/mysql/manifests/server.pp
+++ b/modules/mysql/manifests/server.pp
@@ -81,7 +81,8 @@ class mysql::server ($root_password = '', $debian_sys_maint_password = '') {
     ensure  => file,
     owner   => 'mysql',
     mode    => '0644',
-    require => [User['mysql'], Package['mysql-server']],
+    before  => Package['mysql-server'],
+    require => User['mysql'],
   }
 
   mysql::user { 'debian-sys-maint@localhost':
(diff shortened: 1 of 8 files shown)


tomaszdurka:
Tests are failing on bipbip? is this related?

modules/mysql/manifests/server.pp
```diff
@@ -77,6 +77,13 @@
     notify  => Service['mysql'],
   }
 
+  file { '/var/log/mysql.err':
+    ensure  => file,
+    owner   => 'mysql',
+    mode    => '0644',
+    require => [User['mysql'], Package['mysql-server']],
```
tomaszdurka: Didn't we have a rule to place all the files before package is installed and use flag to keep configuration files on install? This was good approach as the initial run was using the correct files arlready.

modules/mysql/manifests/server.pp
```diff
@@ -109,6 +116,11 @@
     ]
   }
 
+  logrotate::entry{ 'mysql-server-error':
```
tomaszdurka: Formatting

tomaszdurka:
Review done.

modules/mysql/manifests/server.pp
ppp0: Absolutely! This is a different file, though (see https://github.com/cargomedia/puppet-packages/pull/926#issuecomment-129465778). And the aim is just to manage the rights, not the file itself. Bipbip can't access it otherwise (`640` being the original mask). Also, I do not notify `mysql-server` on file change to avoid our servers restarting when deploying this change. Such a thing would have to be orchestrated (master first) and it is not necessary - I tested it: as long as the owner is correct (`mysql`), `mysql-server` won't bother about the error log having opened up a little bit

ppp0:
@tomaszdurka please re-review

modules/mysql/manifests/server.pp
ppp0: Though I'm changing the settings in `my.cnf` so well, the change can't happen unattended

tomaszdurka: Ok then.

tomaszdurka:
Looks good - still the build.

ppp0:
@tomaszdurka I finally changed to `mysql.err` being required by Package - which is consistent with the others. Plus I was mistaken in thinking a service restart wouldn't be necessary. It will be because of the change in log file. So after merging, I'll need to carefully deploy this.

To overcome the annoying build failure (obviously not happening locally :smile: ) I had to put in that silly exec in the spec...

hope you agree, please review

tomaszdurka:
Looks good!

ppp0:
Before merging:
- [ ] stop puppet on db3, db4, db5
- [ ] merge this
- [ ] wait 1-2 min.
- [ ] halt the app (deploy:stop)
- [ ] puppet apply on db3 (will stop master)
- [ ] puppet apply on db4,db5 (slaves)
- [ ] check replication
- [ ] start the app (deploy:start)
- [ ] final check, check correctness of error log
- [ ] start puppet  on db3,db4,db5

ads2 and denkmal should do fine unattended (no master/slave topology)

@cargomedia/devops fyi

PR #930, opened 2015-09-08
Nginx: Enable SSL session cache

njam:
Seems to improve performance additionally to http keep-alive.
```
ssl_session_cache shared:SSL:10m;
ssl_session_timeout  10m;
```

See http://nginx.org/en/docs/http/configuring_https_servers.html


njam:
@kris-lab please review

kris-lab:
makes sense. lgtm

PR #934, opened 2015-09-17
Gearman server should listen on 0.0.0.0 by default

kris-lab:


kris-lab:
@njam could you verify? I am not sure is it good idea to bind to `0.0.0.0` by default. We do it e.g. in mongo. This approach will keep puppet-car*** untouched.

njam:
sgtm, lgtm

PR #935, opened 2015-09-21
Gearman: PHP client fails to connect when listening on "0.0.0.0"

kris-lab:
Listening on 0.0.0.0 was introduced here: #934

Unfortunately with this setting the Gearman php extension can't connect to the server any more (it halts at `GearmanClient::addServer()`.

@kris-lab can you investigate - what the correct setting is? Both on the server and the client?
This is currently affecting developer VMs.

cc @christopheschwyzer

### DIFF
--- a/modules/gearman/manifests/server.pp
+++ b/modules/gearman/manifests/server.pp
@@ -6,7 +6,7 @@ class gearman::server(
   $mysql_password = 'gearman',
   $mysql_db = 'gearman',
   $mysql_table = 'gearman_queue',
-  $bind_ip = '',
+  $bind_ip = undef,
   $jobretries = 25,
 ) {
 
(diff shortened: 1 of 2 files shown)


kris-lab:
The problem is in gearman `--listen` parameter. So far we used `--listen=localhost` what allows to PHP to connect or via `127.0.0.1` or `localhost` (which is/was used in production and development config file). For example if gearman runs with `--listen=127.0.0.1` then PHP can connect only with `127.0.0.1` means `localhost` will not work.

As mentioned in docu if we wanna use all IPs then the best is not to pass the `--listen` param.

@njam wdyt? 
- we can or updated config files with `127.0.0.1` instead of `localhost` and keep `--listen=0.0.0.0` like it is now
- or we can merge proposed code or similar?

njam:
- Default to `undef`
- Add spec to cover this case?

Otherwise agree

vogdb:
@njam what is exactly the recipe to fix it on dev machines now?

kris-lab:
@vogdb @christopheschwyzer could you confirm it works?

vogdb:
@kris-lab confirm

vogdb:
that it works fine now

PR #939, opened 2015-09-29
Remove redis php extension

zazabe:
Switched to predis lib instead, see cargomedia/cm#1923


zazabe:
@njam please review

njam:
lgtm, but merge this only 1-2 weeks after we did the change

zazabe:
@njam could we merge it? 
Anyhow, I don't think that puppet will actually remove the extension, isn't it?

njam:
It will not remove it. But I would like to wait until I've upgraded CM in https://github.com/denkmal/denkmal.org/ - otherwise new installs will fail there.

PR #941, opened 2015-09-30
Pin net-ssh to 2.8.0

kris-lab:


kris-lab:
Instead of using `~` I will stick to `2.8.0` as it is only release before `2.9.x`.

kris-lab:
@njam @ppp0 self merging

PR #942, opened 2015-10-02
nginx: implement reverse-proxy resource

kris-lab:
TODO
- [x] skeleton
- [x] verify proxy works fine
- [x] verify SSL works fine
- [x] verify redirects work fine


kris-lab:
@njam @ppp0 coudl you have a look? 

This is tested and works fine. If you wanna do some test I suggest:
- `rake test:cm:reverse_proxy`
- update `/etc/nginx/conf.d/reverse-proxy-backend-upstream.conf` with real "fuboo" IP address
- update `/etc/nginx/conf.d/vhost_autogen.conf` with real domain `proxy_set_header Host 'foo.xxx';`
- restart

Use `curl -kL --proxy '' -v foo.xxx` to load live website.

njam:
- debug_int: unused
- Can there be multiple instances of this "define" on one node? I guess `nginx::resource::upstream` will collide. Maybe add a test for it?
- Should we also add `/server-status` and a bipbip entry? See `cm::services::webserver`
- spec: Is only testing the HTTP->HTTPS redirect. Let's also test the rest of the code (reverse proxying from a web server?)
- spec: Let's also test the HTTP-only case?
- How will the DNS resolution work? Let's say public DNS resolves `www.example.com` to our proxy. Then how will the proxy know the upstream's server IP address (while the hostname should still be `www.example.com`?

kris-lab:
> debug_int: unused

- ok

> an there be multiple instances of this "define" on one node? I guess nginx::resource::upstream will collide. Maybe add a test for it?

- yes, will not work. I will add some `$name` prefix?

> Should we also add /server-status and a bipbip entry? See cm::services::webserver

- bipbip makes sense. `/server-status` will have a look

> spec: Is only testing the HTTP->HTTPS redirect. Let's also test the rest of the code (reverse proxying from a web server?)

- not sure what exactly you mean?

> spec: Let's also test the HTTP-only case?

ok

> How will the DNS resolution work? Let's say public DNS resolves www.example.com to our proxy. Then how will the proxy know the upstream's server IP address (while the hostname should still be www.example.com?

We override `Host` using [proxy_set_header](https://github.com/cargomedia/puppet-packages/pull/942/files#diff-3a276a41044e3479b4fbc37b8ee7ca71R55) which pass the entire request to the IP of remote server. Remote server recognise Host and serves response.

njam:
> spec: Is only testing the HTTP->HTTPS redirect. Let's also test the rest of the code (reverse proxying from a web server?)

Afaics the spec is only testing the HTTP->HTTPS redirect (`its(:stderr) { should match /< Location: https:\/\/#{name}foo\.xxx\// }`). Shouldn't we also test the HTTPS->HTTPS reverse proxying?

njam:
```
1 cm::services::webserver
 n cm::services::upstream_fastcgi
  n cm::vhost_fastcgi($upstream)

1 cm::services::webserver
 n cm::services::upstream_proxy
  n cm::vhost_proxy($upstream)

# Create upstream, webserver out of the box:
cm::vhost_fastcgi($upstream_name = undef)
cm::vhost_proxy($upstream_name = undef)

# Connect to custom defined upstream:
cm::vhost_fastcgi($upstream_name = 'my-upstream')
cm::vhost_proxy($upstream_name = 'my-upstream')
```

kris-lab:
@ppp0 @njam please have a look on refactoring:
- `cm::services::upstream_proxy` and `cm::services::upstream_fastcgi` class has beed introduced
- `cm::services::webserver` has been extracted
- `cm::vhost_proxy` and `cm::vhost_fastcgi` doesn't make sense imo? It would require a lot of refactoring if implemented as `define` as we create multiple `vhosts` which refers to `fastcgi-backend` upstream. Using this approach we would have to pass `upstream` to the `vhost`? If implemented as `class` then it would be just a proxy to the `cm::services::upstream_XXX`? wdyt?

kris-lab:
retest this please

njam:
- cm::services::upstream_proxy: Should `max_fails` be set to `0`? ("The zero value disables the accounting of attempts." - http://nginx.org/en/docs/http/ngx_http_upstream_module.html)
- cm::reverse_proxy: We'll use multiple of those I understand (with multiple `cm::services::upstream_proxy`) to run multiple domains on one machine. So the upstream name needs to be configurable?
- cm::vhost: I would still make the upstream name configurable. We won't use it, but it will make it clearer how the different defines are tied together?
- spec/reverse_proxy/manifest.pp:
 - Add a test that the reverse proxy serves content from the upstream? E.g. create a basic web server, then server content from it?
 - Add a test with multiple `reverse_proxy`s in one node?

ppp0:
>cm::reverse_proxy: We'll use multiple of those I understand (with multiple cm::services::upstream_proxy) to run multiple domains on one machine. So the upstream name needs to be configurable?

I'm not sure if we handle part of this in puppet-cargomedia with this hiera construct?
```JSON
"cm::services::upstream_proxy::members": [
    "192.168.20.10:1443"
  ],

"foo::proxy::vhosts": [
    {
      "domain": "foo",
      "aliases": [
        "www.foo"
      ]
    }
  ],
```

Surely, this does not allow for several `upstream_proxy` backend groups. We thought this would complicate things and it is not a use case at the moment?

njam:
But if we want to run multiple domains on the same proxy machine we'll need it, right?

kris-lab:
yes, if we want run multiple domains using different upstreams then yes:
- we need define `upstreams` independently (as `define`)
- make sure `upstreams` are called/created
- pass `upstream` name to the `vhost`

So far we create `upstream` as the class because all `vhosts`  uses the same `upstream` (hardcoded name)

njam:
Let's make it configurable then!
How about we make the passing of the `upstream_name` optional, and if it is *not* passed we use a defaults and create the appropriate resource automatically?

kris-lab:
@njam @ppp0 could you have a look. There are still missing parts (like specs):
- upstreams are defined as resource
- we have to make sure that upstream is created independely of `vhost` or `reverse_proxy` resource

It gives us a lot of flexibility but require some important changes in `puppet-cargomedia` [see](https://github.com/cargomedia/puppet-cargomedia/pull/896)

njam:
So the names are now:
```
1 cm::services::webserver
 n cm::upstream::fastcgi
  n cm::vhost($upstream)

1 cm::services::webserver
 n cm::upstream::proxy
  n cm::reverse_proxy($upstream)
```

Two important things imo:
- [ ] Auto-include the upstream resources in both `vhost` and `reverse_proxy` if no upstream name is passed. This will allow to keep the simple use case for single-node environments that we have now. wdyt?
- [ ] spec/reverse_proxy/manifest.pp:
 - [ ] Add a test that the reverse proxy serves content from the upstream? E.g. create a basic web server, then server content from it?
 - [ ] Add a test with multiple `reverse_proxy`s in one node?

kris-lab:
@njam @ppp0 
- added auto creation of default upstream if upstream-name not passed to the vhost!
- still we need to cover all cases by specs

kris-lab:
retest this please

modules/cm/manifests/reverse_proxy.pp
```diff
@@ -0,0 +1,57 @@
...
+  $aliases = [],
+  $redirects = undef,
+  $upstream_name = undef
+) {
+
+  include 'cm::services::webserver'
+
+  $upstream = $upstream_name ? { default => $upstream_name, undef => 'reverse-proxy-backend' }
+
+  if ($upstream_name == undef and defined(Cm::Upstream::Fastcgi['reverse-proxy-backend']) == false) {
```
njam: Use `$upstream` instead of repeating the default name?

Generally `$upstream` and `$upstream_name` seem a bit confusing, as the difference in variable name is not reflecting the difference of the meaning. How about `$upstream_name` and `$upstream_name_real` or something?

kris-lab: Agree! will stick to `$upstream_name` and `$upstream_name_real`

njam:
Code:
- cm::services: Remove `cm::upstream::fastcgi`? This will be created when adding a `cm::vhost` I guess?
- cm::vhost: Pass the `$members` to `cm::upstream::fastcgi`? Because this is the place where the value would be defined if we wanted to override it (vhost living on different node than fpm).
- cm::reverse_proxy: Ditto

Specs:
- cm/spec/reverse-proxy/manifest.pp: Move this test also to a sub-directory, so that it has a "label"? E.g. `two-proxies`?
- cm/spec/reverse-proxy: Add a test to cover the HTTP->HTTPS redirect? (like in `cm/spec/vhost`)

Rest looks very good!

kris-lab:
> cm::services: Remove cm::upstream::fastcgi? This will be created when adding a cm::vhost I guess?

Currently there is include of `cm::services::webserver`. I have replaced that with `cm::upstream::fastcgi` to make sure we have the same output configuration. I agree, maybe it is not needed?

> cm::vhost: Pass the $members to cm::upstream::fastcgi? Because this is the place where the value would be defined if we wanted to override it (vhost living on different node than fpm).

Hmm yes but it applies really only when default upstream is created!

kris-lab:
@njam please have a look on last commit. It would allow to pass `members` to the default upstream!

kris-lab:
@ppp0 was there reason you removed `http` -> `https` test like was implemented [here](https://github.com/kris-lab/puppet-packages/blob/42e774dc71de86afd87cc5d1badd98673dc6b69b/modules/cm/spec/reverse_proxy/spec.rb)

ppp0:
We can re-add it - I was too quick. On the other hand all this redirection stuff is tested for vhost (https://github.com/cargomedia/puppet-packages/blob/master/modules/cm/manifests/vhost.pp#L18) and it looks identical to me?

I will re-add it

kris-lab:
@njam last review? see also https://github.com/cargomedia/puppet-cargomedia/pull/896

njam:
lgtm

kris-lab:
Follow up https://github.com/cargomedia/puppet-packages/issues/945

PR #948, opened 2015-10-11
cm::reverse_proxy: Forward original IP to application

kris-lab:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/887
cc @ppp0 

See https://rtcamp.com/tutorials/nginx/forwarding-visitors-real-ip/

We set headers like `X-Real-IP` already in some modules.

### DIFF
--- a/modules/cm/manifests/reverse_proxy.pp
+++ b/modules/cm/manifests/reverse_proxy.pp
@@ -61,6 +61,7 @@ define cm::reverse_proxy(
       location_cfg_append => [
         "proxy_set_header Host '${upstream_opts[header_host]}';",
         'proxy_set_header X-Real-IP $remote_addr;',
+        'proxy_set_header X-Forwarded-For $remote_addr;',
         "proxy_pass ${proto}://${upstream_name_real};",
         'proxy_next_upstream error timeout http_502;'
       ]
--- a/modules/cm/manifests/reverse_proxy.pp
+++ b/modules/cm/manifests/reverse_proxy.pp
@@ -61,7 +61,6 @@ define cm::reverse_proxy(
       location_cfg_append => [
         "proxy_set_header Host '${upstream_opts[header_host]}';",
         'proxy_set_header X-Real-IP $remote_addr;',
-        'proxy_set_header X-Forwarded-For $remote_addr;',
         "proxy_pass ${proto}://${upstream_name_real};",
         'proxy_next_upstream error timeout http_502;'
       ]
--- a/modules/cm/manifests/vhost.pp
+++ b/modules/cm/manifests/vhost.pp
@@ -61,6 +61,11 @@ define cm::vhost(
     ssl_key             => $ssl_key,
     location_cfg_append => [
       'include fastcgi_params;',
+      'set $_real_client_ip $remote_addr;',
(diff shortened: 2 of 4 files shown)


kris-lab:
@njam have a look on PR. Seems to work as expected
```bash
set $_real_client_ip $remote_addr;
if ($http_x_real_ip != "") {
  set $_real_client_ip $http_x_real_ip;
}
fastcgi_param REMOTE_ADDR $_real_client_ip;
```

njam:
- `X-Forwarded-For` has different semantics than `X-Real-IP`, seems it's used to list multiple IPs that forwarded the request (https://en.wikipedia.org/wiki/X-Forwarded-For). I would suggest to *not* set it, because `X-Real-IP` is sufficient for our use case?
- We already have some default proxy configuration in `proxy.conf.erb` (coming from `nginx::params::nx_proxy_set_header`). Do we really need to set those configs again later on?
- Should we add two specs to cover the desired behaviour?
 - That `cm::reverse_proxy` adds the `X-Real-IP` header
 - That `cm::vhost` sets `REMOTE_ADDR` according to the header

kris-lab:
- [ ] confirm, It is not needed.
- [ ] I would keep it. It is used by vhost/location etc?
- [ ] specs can be added

kris-lab:
About specs...

> That cm::reverse_proxy adds the X-Real-IP header

Tricky! The header is added internally and forwarded to the "live" server. It will require to setup 2 servers?

>  That cm::vhost sets REMOTE_ADDR according to the header

Same problem here!

In both cases we have to pass request via proxy server and then check header structure (1st spec) and header+response (2nd spec). It's a bit complicated imo and tests internal nginx logic?

ppp0:
Spec tests - implementation ideas

>> That cm::reverse_proxy adds the X-Real-IP header

>Tricky! The header is added internally and forwarded to the "live" server. It will require to setup 2 servers?

May be add a second IP address to interface and let curl fire the request from that one. (with e.g. `curl --proxy '' --interface eth0:0 `) Then check that this was the real address

>  That cm::vhost sets REMOTE_ADDR according to the header

One can `exec {}` a `nc` process to listen on a backend socket (e.g. 9001) and pattern match its output (fcgi protocol headers)

kris-lab:
> That cm::reverse_proxy adds the X-Real-IP header

This could be maybe tested by faking the header `curl -H 'X-Real-IP: 1.2.3.4' example.com` ?

> That cm::vhost sets REMOTE_ADDR according to the header

This can be maybe done like you suggest or by configuring vhost to (cgi) with perl/ruby/? So the script would return the IP?

ppp0:
@kris-lab I pushed my experiments. Be aware that I'd like to be able to run `nc` in the background (so that the `exec` resource does not wait for exit) and I did not succeed yet. Second ip address is working well

kris-lab:
I see 2 options:
- run `nc` in spec.rb (not manifest) before we run `curl` e.g. something like that `nc -l -p 9001 >/var/log/spec_nc_log  & curl --proxy '' --interface eth0:0 -m 1 -kL https://foo.xx` ?
- or instead of `nc` create nginx vhost with CGI to e.g. perl/ruby script which would return the IP via HTTP to the curl?

PR #949, opened 2015-10-15
Upgrade MMS backup agent

kris-lab:


kris-lab:
@njam @ppp0 fyi

self merging

PR #953, opened 2015-10-22
Upgrade cm-wowza to 0.2.1

tomaszdurka:


njam:
@tomaszdurka status?

tomaszdurka:
Seems it looks good.
Release is there https://github.com/cargomedia/CM-wowza/releases/tag/v0.2.1.
Plus I understand it has been tested https://github.com/tomaszdurka/puppet-packages/blob/issue-953/modules/wowza/spec/app-cm/spec.rb (can't get CI result).

PR #954, opened 2015-10-23
Add browserify cli support

zazabe:


zazabe:
@njam please review

njam:
The spec is failing (wrong module name?)

Let's also add a spec to ensure the command line tool is available and working, like:
```
  describe command('autoprefixer --version') do
    its(:exit_status) { should eq 0 }
    its(:stdout) { should match 'autoprefixer' }
  end
```

ppp0:
to me, it looks very good!

njam:
I recommended not to test for a specific version because then we need to update the tests when we upgrade the version… :D
lgtm

zazabe:
@ppp0 i don't have write access on this repo, could you merge this PR?

PR #955, opened 2015-10-23
Update mongodb php driver to use "legacy" url

njam:


njam:
@zazabe @alexispeter please review

…they renamed the repo

alexispeter:
I think they created an entirely new driver, which means we should ultimately change to the new driver.

changes lgtm

zazabe:
the new driver is still experimental, it's better to stick with the legacy one, btw you changes pass the test so lgtm!

PR #961, opened 2015-11-04
Add support for puppet master environments

njam:


njam:
@ppp0 please review

Adding "environment" support to puppet, can be useful even if we don't use it now…
Plus some other related improvements.

ppp0:
I haven't tested but it looks good!

PR #964, opened 2015-11-06
Update librarian-puppet to 2.2.1

njam:
Closes https://github.com/cargomedia/puppet-cargomedia/issues/927


njam:
@kris-lab please review

njam:
Tested with a clean puppet master VM, works exactly the same like the previous version but has the *rsync* option!

kris-lab:
lgtm!

PR #951, opened 2015-11-09
Janus webRTC gateway module

ppp0:
Installation
- Janus gateway (https://github.com/meetecho/janus-gateway) / modular-transport branch
- Daemon https://janus.conf.meetecho.com/docs/service.html
- Monit
- Bipbip

Compiling
- we need "--enable-post-processing"

Configuration:
- janus core
- plugins class
- transport class


modules/janus/manifests/deps/libsrtp.pp
```diff
@@ -0,0 +1,11 @@
+class janus::deps::libsrtp(
+  $version = '1.5.0',
+  $build_tests = false,
+) {
+
+  helper::script { 'install libsrtp':
+    content => template("${module_name}/deps/libsrtp_install.sh"),
+    unless  => " ${version}$'",
```
ppp0: not clear how to check the version.. :/ - May be just omit the param?

kris-lab: What about `pkg-config --modversion libsrtp`

ppp0: Works if one has `pkg-config` installed BEFORE building the library. Thank you

ppp0:
Hm I am not completely happy with the specs.. How to properly validate `wss://` ?
@kris-lab @njam @tomaszdurka ideas?

kris-lab:
Just a few thoughts:
- janus: I would not support SSL for now?
- janus: make admin-port, websocket-port + bind interface configurable?
- janus: make token, secret configurable
- janus: make turn/stun server configurable?
- janus: make plugins enable/disable configurable?
- janus::service: add bipbip ?
- janus:plugin: maybe we should add some plugins abstraction for our custom plugins and janus one?

@ppp0 let's pair review on this?

kris-lab:
Also maybe all dependencies should be as separate modules? I mean `srtp`, `usrsctp`, `websockets`?

ppp0:
About the modules: right now, the dependencies are separate classes (`libsrtp`, `libusrsctp`, `libwebsockets`).. If we go make modules out of them we should be aware that they build and provide libraries for further compilation, do may be make them a subclass of  `build` similar to `build::dev::zlib1g` ? Not sure what is more consistent. Otherwise leave it as dep fro Janus as long as they're not depended upon anywhere else

kris-lab:
Agree on dependencies. I was following `x264`, `yasm` etc. but there are binaries apps and not libs only. Let's keep it in Janus then!

kris-lab:
As discussed with @ppp0 
- we should stick to janus `modular-transport` branch
- we should remove SSL as not needed (use default provided by Janus)
- abstract `transport` into classes?
- abstract `plugin` into classes? Disable current plugins! Enable custom plugins

plus copying myself
> janus: I would not support SSL for now?
janus: make admin-port, websocket-port + bind interface configurable?
janus: make token, secret configurable
janus: make turn/stun server configurable?
janus: make plugins enable/disable configurable?
janus::service: add bipbip ?
janus:plugin: maybe we should add some plugins abstraction for our custom plugins and janus one?

ppp0:
>- we should stick to janus `modular-transport` branch

OK

>- we should remove SSL as not needed (use default provided by Janus)

OK

>- abstract `transport` into classes?

OK

>- abstract `plugin` into classes? Disable current plugins! Enable custom plugins

OK - current plugins almost completely deconfigured from the build

> janus: I would not support SSL for now?

OK - support is built-in - just need to provide certificate+key

>janus: make admin-port, websocket-port + bind interface configurable?

OK (see transport configurables)

>janus: make token, secret configurable

to be done

>janus: make turn/stun server configurable?

to be done

> janus: make plugins enable/disable configurable?

as-of-now one can disable plugins in the main config fro janus

>janus::service: add bipbip ?

good idea

>janus:plugin: maybe we should add some plugins abstraction for our custom plugins and janus one?

only our custom plugins at the moment but ready to be expanded if needed

ppp0:
>janus: make token, secret configurable

OK
>janus: make turn/stun server configurable?

OK
>janus::service: add bipbip ?

May be with a plugin? https://github.com/cargomedia/bipbip/issues/132

Otherwise, @kris-lab please review

ppp0:
Hm, we need to write better spec tests for plugin building and installation

kris-lab:
- audioroom: config file should be update; it is not empty anymore [see](https://github.com/cargomedia/janus-gateway-audioroom/blob/master/conf/janus.plugin.cm.audioroom.cfg.sample)
- plugin:audioroom : params should be adjusted according to the config file
- templates:deps:plugin:libinstall.sh : is it good place and name for plugin installer? wasn't it better to move it into `templates/install-plugin.sh`

rest look good

ppp0:
* Config for audioroom is updated
* Agree with refactoring plugin building (see also janus_version enforcement)
* Centralised janus version number

@kris-lab please re-review

kris-lab:
lgtm!

ppp0:
thanks for pair-reviewing!

PR #983, opened 2015-11-11
Deprecation warnings f*** with json output

njam:
@tomaszdurka please review


tomaszdurka:
Looks good

njam:
There was one more "should" in the mongodb package.
Will merge now, so I can run all specs again… Specs succeed locally!

PR #987, opened 2015-11-12
Beautify supported OS output and vm name

ppp0:
Pedantic but friendlier on the eyes and easy to accomplish

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,7 +19,7 @@ Vagrant.configure('2') do |config|
     puppet.module_path = 'modules'
   end
 
-  config.vm.define 'Debian-7', primary: true do |wheezy|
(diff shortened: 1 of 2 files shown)


ppp0:
@tomaszdurka please review :)

njam:
Nooooo - we deliberately removed similar code because it adds unnecessary extra complexity. Especially with a switch/case…
Let's focus on more meaningful things instead!

ppp0:
ok I will shut my eyes when spec testing

njam:
```ruby
ruby -e 'system("say \"0 failures\"")'
```

ppp0:
Now you introduce platform-specific code!

njam:
`\a` doesn't work on OSX :(

PR #989, opened 2015-11-12
nodejs v0.12

ppp0:
add source by doing what this script does:
```sh
# Note the new setup script name for Node.js v0.12
curl -sL https://deb.nodesource.com/setup_0.12 | sudo bash -

# Then install with:
sudo apt-get install -y nodejs
```


tomaszdurka:
Generally looks I would not create this new define in `apt::source` namespace. Let's inline this into `node` package.

tomaszdurka:
@njam please review

njam:
retest this please

tomaszdurka:
good to merge?

njam:
- apt/manifests/update.pp: apt-transport-https: Shouldn't this be in the main apt resource? (update is only to update the index)

rest lgtm

tomaszdurka:
This is not so simple. I vote for keeping this in the apt-update for now (ever?). 
I have dep cycles, after resolving them I still have them. We would need to change all `require 'apt'`` into `include`. 
@njam wdyt?

njam:
Ah, because installing a package requires `apt::update`.
Could we `include` a new resource `apt::transport_https` from `apt` instead?

If not, let's leave it…

tomaszdurka:
@njam please review

I ain't sure about all of this. I could not reproduce https transport problem anymore.
Anyway if there is need for https transport and it would not be installed, I think it would fail on real machines. The only reason it doesn't is because of `before_spec.pp` which requires `apt` and installs missing `apt-transport-https`.

njam:
Sounds a bit risky. What happens if you disable/empty  `spec_before.pp`?

Without HTTPS is not possible?

ppp0:
I disabled `spec_before`'s `apt::update` re-did the tests on a virgin box and it's working.
The package is there:
```
ii  apt-transport-https           0.9.7.9+deb7u7      amd64               https download transport for APT
```

I would assume that it would install on a "real" machine, too, in that case. Ready to be merge, imo

PR #992, opened 2015-11-16
Apt transport https

ppp0:
Follow-up of https://github.com/cargomedia/puppet-packages/pull/989

### DIFF
--- a/modules/apt/manifests/init.pp
+++ b/modules/apt/manifests/init.pp
@@ -1,6 +1,7 @@
 class apt {
 
   include 'apt::update'
+  include 'apt::transport_https'
 
   file { '/etc/apt/sources.list.d/':
     ensure  => directory,
--- a/modules/apt/manifests/transport_https.pp
+++ b/modules/apt/manifests/transport_https.pp
@@ -1,0 +1,5 @@
+class apt::transport_https {
+
+  package {'apt-transport-https':}
+
+}
--- a/modules/apt/manifests/update.pp
+++ b/modules/apt/manifests/update.pp
@@ -26,5 +26,5 @@ class apt::update(
     refreshonly => $refreshonly,
   }
 
-  Exec['apt_update'] -> Package <| |>
+  Exec['apt_update'] -> Package <| title != 'apt-transport-https' |>
 }
--- a/modules/nodejs/manifests/init.pp
+++ b/modules/nodejs/manifests/init.pp
@@ -1,21 +1,22 @@
-class nodejs(
-  $npm_version = '1.4.10'
-) {
+class nodejs {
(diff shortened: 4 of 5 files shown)


ppp0:
@njam you will not like it :)
@tomaszdurka you  probably won't either...

It's a way to break the dep cycle, though... Your thoughts please... I wouldn't spend too much time on this topic, if possible

njam:
Could/should we ignore all resources that *start* with `apt`?

ppp0:
Unfortunately, it does not sound like we could match substrings cf https://docs.puppetlabs.com/puppet/latest/reference/lang_collectors.html#equality-search...

`apt-transport-https` for a start?

tomaszdurka:
Btw, is it ok to install `apt-transport-https` before first running usual `apt-update`?
Maybe it should even:
1. run `apt-update`
2. install `apt-transport-https`
3. run `apt-update`
4. install other packages

?

@ppp0 so `!~` cannot be used there?

ppp0:
May be the first `apt-update` (which is being invoked in `provision.pp`) can be called as an exec? Right now, there is a `require 'apt::update'` which contains this `Exec -> Package` relationship. This `https` tranport is used for nodesource at the moment. I have seen it in another location in the wild as well. I imagine it will become more and more common with https becoming de-facto standard everywhere..

We actually could make it default in our vagrant-box (or even Preseed) and in pulsar-cargomedia's `install_basics` (https://github.com/cargomedia/pulsar-conf-cargomedia/blob/master/recipes/puppet/puppet.rb#L46)

I guess this would be the cleanest option? wdyt?

ppp0:
This is super annoying: after merging, all nodes having a `nodejs` dependency fail with:
```
apt-get -y install  apt-transport-https
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following NEW packages will be installed:
  apt-transport-https
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 109 kB of archives.
After this operation, 167 kB of additional disk space will be used.
WARNING: The following packages cannot be authenticated!
  apt-transport-https
E: There are problems and -y was used without --force-yes
```

The solution, to save time, was: (done with `pulsar` everywhere)
```
monit stop puppet
rm -f /etc/apt/sources.list.d/nodesource.list 
apt-get update
apt-get -y install apt-transport-https
monit start puppet
```

Somehow, adding the `nodesource` list messes up with the apt keys.. See http://unix.stackexchange.com/questions/175764/untrusted-package-apt-transport-https and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=756614 (likely not fully related)

PR #996, opened 2015-11-17
Improve collecting packages by using before

tomaszdurka:


njam:
lgtm

ppp0:
this makes much more sense indeed!

PR #988, opened 2015-11-18
Refactor/rename cm::services::webserver

ppp0:
what this class does is very useful, but not a webserver proper

I propose `cm::services::server_status`

needs a couple of refactorings here and there


ppp0:
@njam please review

njam:
Shouldn't this be the other way around?
- vhost resource should require `nginx` (probably already does?)
- nginx resource should require `nginx::bipbip_entry`

or are there cases where we want nginx *without* the bipbip entry?

njam:
@ppp0 Wdyt?

ppp0:
@njam please review

njam:
lgtm

PR #1003, opened 2015-11-20
Add git::repository resource

njam:
@tomaszdurka please review

A resource to create a checkout of a git repo at a specific tag.
Can be used for cm-janus modules (#999)

cc @ppp0


tomaszdurka:
Mixed feelings about name. Maybe `git::repo(sitory)`? I believe you have thought about it. On the other hand the way you've put this in description makes sense.
Isn't this a git repo with specific "ref state"?

Btw you are testing files and their contents. Have you thought about using git commands rather than this?

njam:
I like `git::repository($remote)`.

I thought about using git commands in the specs. But it will end up being almost the same as in the manifest, which might be a bit risky, if there's a bug…

@tomaszdurka wdyt?

njam:
@ppp0 please review

tomaszdurka:
Looks good. Maybe I would rather use `ref` instead of `version`?

njam:
@tomaszdurka please review

tomaszdurka:
Looks good.

PR #997, opened 2015-11-20
Don't run apt::update before actual spec manifest

tomaszdurka:
I understand we rely that some packages are present on the system or at least are within the manifests we run when using those packages. That's why we have `spec_before.pp` which basically prepares ground for specs by installing required modules. *(I don't agree with this, I think modules themselves should require dependencies and not rely they are included elsewhere - but let's ignore that for now).

Currently we apply `spec_before.pp` which adds some prerequisites like `apt`, apt-repo is up to date. This sounds reasonable, but problem is this manifest is applied in separate puppet apply run here in tests, while it runs together with modules from spec in production. Possible errors - like last one with nodejs.

Steps to take:
1. drop `apt::update` in `provision.pp` (or only, manually run `apt-get update` command) *# I would drop it fully*
2. run spec_before's `apt::update` together with spec manifest, so either:
  - require 'apt-get' in each spec manifest needing it
  - merge manifests (spec_before + spec's manifest) and run together - tested to be working


tomaszdurka:
```
Failure! 199 specs run, 3 failures (826 examples, 10 failures)
Failed examples:

    apt::update Command "stat -c "%Y" /var/lib/apt/periodic/update-success-stamp" should not update indices
    puppet::master with passenger Package "passenger" 
    puppet::master with passenger Package "apache2" 
    puppet::master with passenger Port "8139" 
    puppet::master with passenger Port "8141" 
    passenger Package "passenger" 
    passenger File "/etc/apache2/mods-enabled/passenger.load" 
    passenger File "/etc/apache2/mods-available/passenger.conf" 
    passenger Port "80" 
    passenger Command "sudo apache2ctl -M" stdout 
```

tomaszdurka:
I've run `puppet::master with passenger Package "passenger" ` after destroying the box and didn't get this error. Is this possible that current apt-get state (from vbox on CI) doesn't reflect current sources?

tomaszdurka:
Retest this please.

njam:
Yes, vagrant box probably still has backports. But I don't see how it should affect puppet (which is installed from a separate source). Maybe it's flaky?
I you can't reproduce it I would suggest to fix `apt::update` and merge it?

njam:
We should include `apt::update` (or `package_manager`) whenever we install packages.

tomaszdurka:
Still one failing: `apt:cache-valid`

tomaszdurka:
@njam please review

tomaszdurka:
@ppp0 please review

(still missing Puppetfile additions)

njam:
- pulsar_rest_api/manifests/init.pp: Doesn't need apt?
- rsyslog/manifests/init.pp: Needs apt

tomaszdurka:
@ppp0 please check

I've wrote small rake task to add apt mod to `Puppetfile` for all modules which does have `require 'apt'` within it's manifests. But only to those which doesn't have it yet. 
There should be no misses - nor duplicate declarations.

tomaszdurka:
@ppp0 please review

tomaszdurka:
Discussed with @ppp0 - merging.

PR #999, opened 2015-11-25
janus modules: adjust "unless" to consider version

ppp0:
Janus and janus plugins are installed from git sources. We install a specific sha, but don't check it in the `unless`.

Idea: Use git::repository (https://github.com/cargomedia/puppet-packages/pull/1003)

Maybe extract that into a `git::checkout` resource or similar?
cc @tomaszdurka

### DIFF
--- a/modules/janus/manifests/deps/libsrtp.pp
+++ b/modules/janus/manifests/deps/libsrtp.pp
@@ -1,5 +1,5 @@
 class janus::deps::libsrtp(
-  $version = '1.5.0',
+  $version,
 ) {
 
   require 'build::pkg_config'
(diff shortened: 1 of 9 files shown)


ppp0:
(specs expected to be failing)

ppp0:
could [this](https://github.com/cargomedia/debian-packages/issues/16) be a way to handle versioning in a simpler way? (https://github.com/cargomedia/debian-packages/issues/16) @cargomedia/devops thoughts?

njam:
Yes, agree, I think building our own debs would make sense.

ppp0:
obsoleted by https://github.com/cargomedia/debian-packages/pull/16

njam:
Followup to adjust modules: https://github.com/cargomedia/puppet-packages/issues/1035

PR #1008, opened 2015-11-26
Upgrade socket-redis to "latest"

njam:


njam:
Paired with @tomaszdurka

PR #1012, opened 2015-11-27
Remove duplicate declaration from provision script (immediate solution)

tomaszdurka:


tomaszdurka:
@njam please review

njam:
Let's keep it but rename it, so we'll remember that we need to have `apt-get update` before provisioning

njam:
ok… :D

njam:
Killing jenkins build, because they use up all our workers.
I will merge this, let's wait for the nightly build!

PR #1016, opened 2015-12-01
First successful build on ubuntu 1504, git and polipo module

ppp0:
Goal: `bundle exec rake spec:git` to complete successfully

Needs some adaptations to e.g. puppet apt sources, polipo, what not

PR #1018, opened 2015-12-01
Follow-up: fix specs

ppp0:
From #1016


ppp0:
Specs are running on CI but they fail miserably
```
==> Ubuntu-1504: Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install ri' returned 100: Reading package lists...
      ==> Ubuntu-1504: Building dependency tree...
      ==> Ubuntu-1504: Reading state information...
      ==> Ubuntu-1504: Package ri is not available, but is referred to by another package.
      ==> Ubuntu-1504: This may mean that the package is missing, has been obsoleted, or
      ==> Ubuntu-1504: is only available from another source
      ==> Ubuntu-1504: 
      ==> Ubuntu-1504: E: Package 'ri' has no installation candidate
      ==> Ubuntu-1504: Error: /Stage[main]/Ruby/Package[ri]/ensure: change from purged to present failed: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install ri' returned 100: Reading package lists...
```

And I know this is gonna be fixed by a `apt-get update`
@tomaszdurka any ideas?

tomaszdurka:
Where are you getting this message? Locally? Here I can see some other errors about virtualbox.

njam:
@ppp0 to me it looks like this package is not available on ubuntu.
```
      ==> Ubuntu-1504: E: Package 'ri' has no installation candidate
```

Probably the specs never succeeded on both platforms?

ppp0:
Ok, so part of the problems is that Ubuntu 15.04 now has ruby2.1 as default (vs 1.9 before) and this changes how rubygems is installed (the `gem` binary).. I propose some changes as pushed - needless to say that it works on my local vm

njam:
- ruby/manifests/init.pp: Shouldn't we install a "ruby"? http://packages.ubuntu.com/xenial/rubygems
- Specs are failing because of polipo?

ppp0:
* not sure.. I'll test this makes sense ([vivid version of the above](http://packages.ubuntu.com/vivid/rubygems))
* yes, I'm on it.. it's because we removed `universe` from https://github.com/ppp0/puppet-packages/blob/578c3ac5aeb3d659dd5bb6cc202896cc65144cc9/modules/apt/templates/sources-vivid

njam:
ruby/manifests/init.pp: Just that this puppet module is called "ruby", so I'd say we should install the "ruby" apt package. This in turn *provides* "rubygems". wdyt?

ppp0:
fully agree, makes sense - now waiting for ci to complete, looks good

@njam you want to review ?

njam:
What's the reason to not install "ruby-dev" on ubuntu? I see they have it too.
I think we should either install it, or not. If it's already there I suggest we keep it. Not sure why it is there though…

ppp0:
I kinda remember it was because of issues with [installing this gem, Nokogiri](http://www.nokogiri.org/tutorials/installing_nokogiri.html) which is now expected (at least tests seem to confirm that) that it should  Just Work™

njam:
Ok then, fingers crossed! lgtm.

ppp0:
merging with my left hand ;)

PR #1031, opened 2015-12-09
Changes to Vagrantfile (forwarded ports)

ppp0:
Due to forwarded ports being persisted when snapshotting, on CI there might be port collisions from reviving snapshots vs freshly upped instances. Solution: hardcode forwarded port high enough not to interfere with vagrant's counting scheme


ppp0:
@njam please review

njam:
lgtm

PR #1013, opened 2015-12-11
cm-janus: changes on config file should restart service

tomaszdurka:
Same for janus.


tomaszdurka:
Paired with @dergraf

@njam please review

njam:
lgtm!

PR #1038, opened 2015-12-12
Use a fact "bootstrap" for bootstrapping puppet

### LINKED ISSUE #721: Fix puppet provision step
Tag-based provisioning doesn't work reliably, because no dependencies are resolved.
Usage in [pulsar-conf-cargomedia's puppet.rb](https://github.com/cargomedia/pulsar-conf-cargomedia/blob/master/recipes/puppet/puppet.rb#L16).
@kris-lab @ppp0 can you look into this? cc @tomaszdurka 

Idea:
Using (directory) environments, then selecting a *special* environment `provisioning` for the first run.
https://docs.puppetlabs.com/puppet/latest/reference/environments.html
https://docs.puppetlabs.com/puppet/latest/reference/environments_assigning.html
https://docs.puppetlabs.com/puppet/latest/reference/environments_creating.html

But can we easily reuse all modules for the new environment?

njam:
Paired with @kris-lab
Closes #721

### DIFF
--- a/modules/puppetmaster/manifests/init.pp
+++ b/modules/puppetmaster/manifests/init.pp
@@ -7,7 +7,6 @@ class puppetmaster (
   $puppetdb_port_ssl = 8081,
   $bootstrap_classes = [
     'puppet::agent',
-    'monit' # See https://github.com/cargomedia/puppet-packages/issues/232
   ],
   $puppetfile = undef,
   $port_webrick = 8140,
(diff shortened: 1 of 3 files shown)


njam:
@kris-lab please double check

kris-lab:
lgtm

njam:
@tomaszdurka @ppp0 fyi

tomaszdurka:
Looks good. Makes more sense. Did you think about rather doing bootstrap when no `bootstrapped` fact? We could add this fact after bootstrapping. Current allows bootstrapping multiple times - plus?

njam:
Could be as well! Doesn't make much of a difference imo.

PR #1066, opened 2015-12-14
Comment out monit-deps that lead to dependency cycle

njam:


njam:
@kris-lab please review

PR #1068, opened 2015-12-14
Upgrade to puppet 4

njam:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/817

Docu about path changes:
https://docs.puppetlabs.com/pe/latest/install_upgrading_path_changes.html#new-codedir-holds-all-modulesmanifestsdata

We should wait for version 4.4 (january/february) due to this bug:
https://tickets.puppetlabs.com/browse/PUP-5465

TODO:
- [x] Update puppet agent
- [x] Update puppetmaster
- [x] Update puppet-cargomedia
 - [x] Usages / sandbox
 - [x] specs
- [x] Update vagrant-boxes (scripts/)
- [x] Update specs (spec/lib/puppet.rb)
- [x] Bipbip https://github.com/cargomedia/bipbip/pull/146
- [x] Run all specs with puppet 4
- [ ] New vagrant-boxes
- [x] Delete passenger module
- [x] pulsar-conf-cargomedia


njam:
@kris-lab @ppp0 please review
Everything is ready for puppet4! The following is a short summary of the changes:

Changes regarding the puppet modules for "puppet" and "puppetmaster/server":
- Upgrade from puppet3 to puppet4 (duh)
- Rename *puppetmaster* to *puppetserver* (it's a java application now, no more apache)
- Many directories and file paths have changes (see [Where did everything go?](https://docs.puppet.com/puppet/4.4/reference/whered_it_go.html))
- Binaries are *not* on the PATH anymore by default (but in `/opt/puppetlabs/bin/`). I'm symlinking them to `/usr/local/bin` for convenience.

Changes on other modules:
- Mostly minor changes regarding problems in some modules' code that doesn't work anymore in puppet4
 - For example there's a warning when accessing undefined variables. So instead of accessing `$::copperegg_tags` I'm now accessing `$::facts['copperegg_tags']`.

Changes regarding the development VMs:
- Using the `-plain` flavors of our VMs (which don't have puppet installed yet), and then installing puppet with our `scripts/puppet-install.sh` script. This allows to install puppet4 into an previously empty box.

njam:
(note: the specs fail because CI has VM snapshots with puppet 3. I've run *all* specs locally with puppet 4 and everything works fine.)

kris-lab:
In general all looks good. However this change is pretty big and almost impossible to review!

I suggest we deploy and see what happens? :)

njam:
ok let's do that! I will deploy it tomorrow. We will need to adjust all vagrant boxes that use puppet provisioning. So I'll do that after it's deployed to production. Wish me luck.

PR #1069, opened 2015-12-14
Enforce puppet "service"'s "provider" with custom fact

njam:
The default "provider" should work fine in the new puppetserver 2, but not now


njam:
Paired with @kris-lab 

@tomaszdurka fyi this is a workaround for the wrong "provider" default attribute for `Service` in our puppet. This is fixed in the newest "puppetserver 2" (which replaces puppetmaster)…

tomaszdurka:
Oh, so we are putting some global things in the end? Not sure if I am against it. I understand it's super quick solution.

Won't we have problems for some modules. When used without this site.pp? E.g. with puppet apply, or within specs?

njam:
Agree, it's suboptimal. Only alternative is to upgrade to puppet 4 afaik. This is currently not possible due to another bug (see https://github.com/cargomedia/puppet-packages/pull/1068).

It will not be applied in specs…

PR #1072, opened 2015-12-14
Re-add "service_provider" for daemon, for puppet-master-less setups

njam:


njam:
Paired with @kris-lab

PR #1074, opened 2015-12-15
Xpra module

kris-lab:


kris-lab:
@njam please have a look

njam:
- manifests/display.pp: Don't allow to configure `--no-daemon`, because it is needed for the module to work?
- manifests/display.pp: How about just `require 'xpra'` here?

kris-lab:
@njam have a look

njam:
lgtm!

PR #1075, opened 2015-12-15
Extend daemon with evn params

kris-lab:


kris-lab:
@njam have a look. Initially I tried to make it more configurable like:
```
  $unit_opts = {},
  $service_opts = {},
  $install_opts = {}
``` 
which would fir nicely into service template. However it would conflict with current argument and also would make it hard to configure with keys like `Environment` which could be used multiple time (array has to be there anyway).

wdyt?

kris-lab:
Hmm I have missed compatibility with `sysvinit`. So we have to keep it simple as possible. Not sure what would be the best pattern for it? `/etc/default/service-name` which contain variables or like it is know with `export` or maybe we can pass it to `start-stop-daemon` if we use `--exec` and `env` prefix?

njam:
- Can we pass a hash instead of an array? (key => value)
- init script: Is `export` really needed?

kris-lab:
- not sure, probably not. We have to pass `Environment` multiple time then?
- `export` not needed imo. Just there will be different behaviour for `systemd` and `sysvinit`. I can drop `export` for sysvinit and workaround check in `spec`

njam:
- Can't we pass `environment => {'FOO' => 'foo1', 'BAR' => 'bar1'}`?
- About `export` - I was just not sure if it's needed to be exported so it's available in the process. The spec is fine!

kris-lab:
- yes, hash makes sense. I was thinking about different thing! sorry...
- `export` is needed to be available in the process, yes!

@njam please review

kris-lab:
@njam added small change

njam:
- sysvinit.sh.erb: Also use foreach? Might be bit easier…

Rest lgtm

kris-lab:
Sorry: `Detail: undefined method `foreach' for {"DISPLAY"=>":99", "FOO"=>"BOO"}:Hash`

Will merge like it is now.

njam:
Ah - I meant to use the same approach for systemd and sysvinit:
```
<% @env.each do |name, value| -%>
```

kris-lab:
Ahh using multiple `export` per `name=value`? Currently in `systemd` is multiline and `sysvinit` in single line. Maybe could be better to have multiline as well. agree.

PR #1076, opened 2015-12-16
Use daemon for cm_janus

tomaszdurka:


tomaszdurka:
@dergraf @ppp0 please review.

As mentioned to Andre I had problems keeping user as `cm-janus:cm-janus`.
To reproduce change user to `cm-janus:cm-janus`.
Strange as according to `/etc/passwd`, `/etc/group` files user and group are present. 
When running daemon with `cm-janus` only it works and in the end it works the same way (takes `cm-janus` as group).

Questions:
- is `user:group` pair working for systemd?
- is there something off with our daemon init script? - it worked with previous script

dergraf:
@ppp0 any ideas regarding this?

ppp0:
Interesting..

I agree that `.conf` files belonging to the same user/group of the process needing them looks kinda good, but I also agree, that this was not the case before. `.conf` files where happily owned by root.

tomaszdurka:
Indeed. Conf files didn't need to be changed. There was problem with log-dir. Could not start it as `cm-janus` user without setting it as owner of the log-dir.

ppp0:
I'm also experiencing issues with logging (see https://github.com/cargomedia/puppet-packages/blob/master/modules/janus/templates/init.sh#L31 - old school redirection to a log file). In systemd this is no problem as logging come included (`journalctl -u janus` e.g.)

njam:
- manifests/init.pp: I think config files should be owned by `root` (security considerations)
- manifests/init.pp: Daemon should require config file
- init.sh: delete

Otherwise this looks good, and could be merged?

tomaszdurka:
@njam please confirm

njam:
lgtm

PR #1080, opened 2015-12-17
Add janus config file override, specs, re-add logging

ppp0:
~~follow-up of #1035~~(reverted) 
As asked for in https://github.com/cargomedia/puppet-packages/pull/1035#issuecomment-165123943

~~Depends on #1078~~

PR #1082, opened 2015-12-18
Optionally compile Janus and plugins from source

dergraf:
The tests for the plugins/transports are currently NOT using a source install.

@kris-lab, @njam, @ppp0


njam:
- janus/manifest/init.pp: When installing from source we will install the version from `janus::version::number`, but never check for updates (the `helper::script`'s `unless` is just `ls /usr/bin/janus`). I would suggest these changes:
 - Combine $version and $use_src into one parameter ($src_revision)
 - Use `git::repository` to check out a git clone of the repo at a specific revision
 - Trigger compilation (which can be `helper::script`, or better just `exec`) whenever the repo changes (use `notify` from `git::repository`)
- janus/manifests/plugin.pp: Same
- I'm not too fond of this `janus::version::number` mechanism - it's kind of magic - and will apply to all resourced related to janus.
 - How about for the plugins having two params: $src_revision, $janus_src_revision and then explicitly passing it along?
- janus/manifests/service.pp: Can we always use `daemon`?
 - The `service`'s `subscribe` should also be present for the non-src case (when using `daemon`)
 - Maybe this resource can be completely inlined into `janus`, which will simplify things further
 - I suggest we **upgrade janus to a version with log file**. Otherwise everything will get overcomplicated. Could you build the appropriate DEBs - then I can test it in a local VM, and if it works fine we can continue here?

njam:
```
    exec { 'compile janus':
      path        => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
      command     => template("${module_name}/install.sh"),
      unless      => 'ls /usr/bin/janus',
      timeout     => 900,
      refreshonly => true,
      subscribe   => Git::Repository['foo'],
    }
```

ppp0:
Also do not forget to 
>Add janus config file override, specs, re-add logging

(from #1080)

ppp0:
paired w/ @dergraf 
@njam please review
@kris-lab cc

dergraf:
yay!

njam:
Looks very good!

- manifests/init.pp: Let's use mode `0644` also for directories - puppet automatically converts to `0755` - makes it easier imo to be consistent (the `cert.pem` should not be executable?).
- How about making those git repo paths more consistent (now `/opt/janus-gateway` and `/opt/janus/build/${name}`). Maybe just `/opt/janus/janus-gateway` and `/opt/janus/janus-gateway-audioroom`?
- plugin/rtpbroadcast.pp: Remove commented out code
- manifests/transport.pp: Remove this file?
- Make those `make install` execs `refreshonly` (so they don't execute every time)
- Installing janus or a plugin should trigger a service refresh (from both source and package installs)

ppp0:
* agree, plus the `cert.key` should much less be world readable
* agree, some refactoring had to be done (introduce var for plugin repo name)
* yes
* see #epicfailure after https://github.com/dergraf/puppet-packages/commit/0ba73a8fc92446441395132925e335a898aaa2fd (had to https://github.com/dergraf/puppet-packages/commit/87ba10b96fc7d775b2ccf3f30ec57f6438512461) - can you suggest what went wrong? (Duplicate declaration)
* ok
* ok

ppp0:
@njam please review (CI will take a while)

njam:
Regarding 0ba73a8fc92446441395132925e335a898aaa2fd: Probably "janus::transport::websockets" should not include "janus" *and* the other way around? I would say transports should be optional/pluggable, so *they* should include janus?

ppp0:
do you mean removing https://github.com/cargomedia/puppet-packages/blob/master/modules/janus/manifests/init.pp#L31 and https://github.com/cargomedia/puppet-packages/blob/master/modules/janus/manifests/init.pp#L32 ?

njam:
yes, wdyt?

ppp0:
sure.. well, that was easy :)

@njam please review

njam:
lgtm

ppp0:
retest this please

ppp0:
@njam seems good now, I'm merging

njam:
lgtm

PR #1079, opened 2015-12-18
Upgrade copperegg "Revealcloud Collector"

njam:
Copperegg informed us that they recommend us to upgrade:

> This is to inform you that we have released a new stable version of Revealcloud Collector, which has many bug fixes.
We found that auto-upgrade of Revealcloud Collector is disabled on some of your servers, which is preventing the new collector version to get installed.

> To enable auto-updates on Linux collector, kindly remove "-x" flag from the command of revealcloud in the startup files
/etc/init/revealcloud.conf 
/etc/init/revealcloud.
Thereafter restart the collector.

I would suggest we upgrade and enable auto updates.
@ppp0 could you look into this?


njam:
@kris-lab please review (maybe it's easier to look at the full code as it's quite a rewrite)

njam:
@ppp0 @tomaszdurka please review  (maybe it's easier to look at the full code as it's quite a rewrite)

ppp0:
lgtm, nice one, being able to check a process args 8` its(:args) { should match '-t fact-tag-foo' }`from `describe process('foo')`

PR #1000, opened 2015-12-22
Add module to blacklist kernel modules

njam:
Add a file to `/etc/modprobe.d/`, which contains 1 or more lines of:
```
blacklist <module name>
```

### DIFF
--- a/modules/kernel/manifests/modprobe.pp
+++ b/modules/kernel/manifests/modprobe.pp
@@ -1,0 +1,13 @@
+define kernel::modprobe(
+  $content
+) {
+
+  file { "/etc/modprobe.d/${title}.conf":
+    ensure  => file,
+    owner   => '0',
+    group   => '0',
+    mode    => '0644',
+    content => $content
+  }
+
+}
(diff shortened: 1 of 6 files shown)


njam:
@ppp0 @kris-lab please review

ppp0:
I wonder if `modprobe` shouldn't be refreshed/called? Or name the define `modprobe_conf` ?

kris-lab:
Looks good. 

In general to apply it fully you should reboot system? Otherwise you have to make sure the specific module is loaded or unloaded from kernel with `modprobe module-name` or `modprobe -r module-name`?

njam:
Ok agree, please review!

kris-lab:
afaics lgtm

njam:
@ppp0 please review

ppp0:
>In general to apply it fully you should reboot system? 

I'm not completely sure on this one - but surely, it gives you a clean, defined state. To me, `modprobe` sounds like the util bin loading the module whereas here we manipulate its config

but be it, lgtm

njam:
Afaik "udev" uses "modprobe" to load modules when devices are attached. Now I'm unloading a module if it's loaded when we blacklist it. So it should be in the desired state.

PR #977, opened 2015-12-23
Kiosk mode Chromium

njam:


njam:
@kris-lab @ppp0 please review

Summary:
- **lightdm**: Managing the display manager, allowing to configure "autologin", allowing to configure custom "xsession" definitions (in our case just running a script which will start chromium).
- **accountsservice (freedesktop)**: There's a daemon called `accounts-daemon` which is part of the freedesktop.org project. All major display managers (including lightdm) use it for account configuration. We can talk to it via "dbus". I've added a resource to configure a `accountsservice::user`, where we can pass a specific user's "xsession" to use.
- **chromium**: Just adding the package, and putting everything together in a `chromium::kiosk`.

I've added a flag to enable virtualbox GUI during specs. You can test everything with:
```
bundle exec rake spec:chromium:kiosk gui=true
```
Also lightdm doesn't start automatically after installation, so you'll need to start it with `systemctl start lightdm`.

ppp0:
lvgtm! Nice one!

I was able to randomly type some ctrl-* (various combinations), open the source view of the page and end chromium (which correctly threw me back to the lightdm login screen)

I think we can live with this, given that there will be no keyboard anyway. On the lightdm login screen one is able to autologin back with hitting a button - good for touch screens

njam:
You can also hit Alt+F4 to achieve the same. I think it's good as it allows us to also log in with a different (administrative) account using the standard ubuntu xsession.

The specs failed because "dbus" didn't yet pick up the new configuration after installing the "policykit-1" and "accountsservice" packages. Unfortunately I can't find a way to synchronously reload dbus' configuration.
So I've added `sleep 1` after installing the package. Let's hope it works on CI…

PR #1084, opened 2015-12-23
Specifically run logrotate as root

njam:
@ppp0 please review

Currently logrotate fails on ubuntu, because log files have the group "syslog" (which doesn't exist on debian).
Enforcing using "root" seems to solve the problem.


ppp0:
oh, ok..
sgtm and lgtm, too

PR #1086, opened 2015-12-23
Only unload kernel modules when blacklist changes

njam:
…otherwise will unload every time.

@kris-lab please review


kris-lab:
Hmm makes sense! Lgtm

PR #952, opened 2015-12-23
coturn module

ppp0:
See https://github.com/coturn/coturn


ppp0:
- Log file could be config'd (default is supposed to be `/var/log/coturn.log`- might need to be managed in puppet, too
- TLS and DTLS disabled in config
- Lots of more options that could be made configurable.. @kris-lab opinion?

ppp0:
CI failing for Ubuntu.. :(  - will check tomorrow

kris-lab:
Looks very good but I think we miss some ACL in this module. 

Janus basic `turnserver` configuration is like [this](https://github.com/meetecho/janus-gateway/blob/master/html/echotest.js#L78) so I think we have to make it configurable somehow? Maybe even per subscriber?

I see there are many ACL option with mysql, mongo etc. We have to decide to one way. @njam @ppp0 wdyt?

ppp0:
@kris-lab @njam whoever wants to review this first!

I took out the authentication (will be properly implemented in #1087)

njam:
@kris-lab could you look into this?

Just one thought: from some discussion threads on the Internet I remember that authentication is required for coturn to work with WebRTC. It can of course be just a single set of "username/credential". If that is the case then we need to enable it and make it configurable.

kris-lab:
##### `WebRTC` and `ICE` and `coturn`

yes, agree, it is required to use authentication

> http://stackoverflow.com/a/26110678
In summary:
- yes, long-term authentication is required for WebRTC to use TURN.
- yes, it seems that you can just hard-code a single set of credentials for everyone to use -- coturn isn't bothered that two clients get allocations simultaneously with the same credentials.
- one possible solution for proper security with minimal hassle would be a TURN REST API, which Coturn supports.

kris-lab:
I have been focused on two aspects:
- `credentials`: there is a huge room for hard-coding `username:password` into the config file but better is store it in `sqlite` or `mysql` or `redis` or `postgreql`. It can be achieved by running CLI tool `turnadmin` from e.g. `puppet` (add/del/list users etc.)
- `statistics`: there is way to receive `stats` for users with `pub/sub` using `redis`. Unfortunately I could not read stats asynchronously. For `bipbip` we need a way to read info rather than receive it?

kris-lab:
The biggest problem is to verify that `ICE` really connects to the `coturn`. I could not find any way to simulate this behaviour in the local network. Still investigating...

kris-lab:
Hardcoded credentials seems to work fine.

#### Generate user
```bash
$ turnadmin -k -u njam -p njam -r njam.com
0x009da43cb799f01739d6d3dabcee6bb6
```
Update config `/etc/turnserver.conf`
```bash
user=njam:0x009da43cb799f01739d6d3dabcee6bb6
realm=njam.com
```

#### Verify auth
```bash
$ turnutils_uclient -u njam -w njam -T 127.0.0.1
0: Total connect time is 1
0: 2 connections are completed
0: start_mclient: msz=2, tot_send_msgs=0, tot_recv_msgs=0, tot_send_bytes ~ 0, tot_recv_bytes ~ 0
1: start_mclient: msz=2, tot_send_msgs=0, tot_recv_msgs=0, tot_send_bytes ~ 0, tot_recv_bytes ~ 0
2: start_mclient: msz=2, tot_send_msgs=0, tot_recv_msgs=0, tot_send_bytes ~ 0, tot_recv_bytes ~ 0
```

kris-lab:
#### Turn server management API
There is option to get some useful information over `telnet`:
````bash
$ telnet localhost 5766
> pu

    user: <njam>, 2 sessions

  Total sessions: 2

> ps

    1) id=128000000000000076, user <njam>:
      realm: njam.com
      started 761 secs ago
      expiring in 16 secs
      client protocol UDP, relay protocol UDP
      client addr 127.0.0.1:43176, server addr 127.0.0.1:3478
      relay addr 127.0.0.1:61038
      fingerprints enforced: ON
      mobile: ON
      usage: rp=3, rb=308, sp=3, sb=364
       rate: r=0, s=0, total=0 (bytes per sec)

  Total sessions: 1
```

njam:
> there is a huge room for hard-coding `username:password` into the config file but better is to store it in `sqlite` or `mysql` or `redis` or `postgreql`.

Why is it better to *not* use the config file? Wouldn't it be easier?

> The biggest problem is to verify that `ICE` really connects to the `coturn`. I could not find any way to simulate this behaviour in the local network. Still investigating...

Agree, not so easy… Maybe we can enforce the type of connection establishing to use in the client? Chrome shows this "iceTransportType = all" in the internals. Maybe we can enforce "relay"?

> There is an option to get some useful information over `telnet`

This looks like exactly what we need! With a telnet client in ruby, should be simple enough.

kris-lab:
> Why is it better to not use the config file? Wouldn't it be easier?

Yes, of course it is easier and currently even better for us. Basically, we decided already to go with hardcoded values in config file. It works OK. I was learning `coturn` so I have found many options... it is pretty nice piece of software. Also `turnserver` creates `/var/db/turndb` by default which is `sqlite`.

##### Overall:
1. I am going to make sure for 100% we can simulate using of `turnserver` by ICE with browser
2. I will hard code credentials and it seems to be ready to go then!
3. I will optimise some more like e.g `realm`, `mobility` etc.
3. I have opened `bipbip` plugin ticket https://github.com/cargomedia/bipbip/issues/137

kris-lab:
follow up @ #1091

kris-lab:
##### Forcing `ICE` to use `relay`
It is doable to force at least `chromium` to use `realy` as `iceTransports`. You can achieve this by:
```js
var iceServers = [{url: "turn:10.10.10.111:3478", username: "njam", credential: "njam"}]
var pc_config = {"iceServers": iceServers, "iceTransports": "relay"};
var pc_constraints = {
  "optional": [{"DtlsSrtpKeyAgreement": true}]
};
var pc = new RTCPeerConnection(pc_config, pc_constraints);
```

`coturn` session description
```bash
> ps

    1) id=132000000000000010, user <njam>:
      realm: njam.com
      started 25 secs ago
      expiring in 575 secs
      client protocol UDP, relay protocol UDP
      client addr 10.10.10.1:51789, server addr 10.10.10.111:3478
      relay addr 10.10.10.111:64024
      fingerprints enforced: OFF
      mobile: OFF
      usage: rp=795, rb=66677, sp=42213, sb=57799001
       rate: r=2778, s=2408291, total=2411069 (bytes per sec)
      peers:
          10.10.10.111
          10.10.10.111:50510
          10.0.2.15

  Total sessions: 1
```

`relay` using pipeline `janus` -> `coturn` -> `browser` and credentials works as expected

cc @njam

kris-lab:
##### Importnat for recrods

I have noticed that setting `coturn` credentials in `janus` config has no effect at all! After investigation of workflow it looks like:

**Step 1** `Browser/ICE` sends request to `turnserver` using credentials and...

**Step 2** `Browser/ICE` receives session description from `turnserver` like:
```
a=ice-ufrag:9XCV
a=ice-pwd:ae/n4fHeVkW1oiJvGtcLeZ
a=ice-options:trickle
(...)
a=candidate:12 1 udp 1006633215 10.10.10.111 56156 typ relay raddr 10.10.10.111 rport 55010
a=candidate:14 1 udp 1006633215 10.10.10.111 53962 typ relay raddr 10.0.2.15 rport 43317
a=candidate:3 2 udp 2013266430 10.10.10.111 46167 typ host
a=candidate:4 2 udp 2013266430 10.0.2.15 59463 typ host
a=candidate:12 2 udp 1006633214 10.10.10.111 62205 typ relay raddr 10.10.10.111 rport 46167
a=candidate:14 2 udp 1006633214 10.10.10.111 65009 typ relay raddr 10.0.2.15 rport 59463
```
and forward it to the `Janus`

**Step 3** `Janus` receives `ICE` candidates from the `Browser` like:
```bash
[658704224] Trickle candidate (audio): candidate:143138880 1 udp 41819903 10.10.10.111 57168 typ relay raddr 0.0.0.0 rport 0 generation 0
[658704224]  Adding remote candidate component:1 stream:1 type:relay 0.0.0.0:0 --> 10.10.10.111:57168
[658704224] ICE already started for this component, setting candidates we have up to now
[658704224] ## Setting remote candidates: stream 1, component 1 (1 in the list)
[658704224] >> Remote Stream #1, Component #1
[658704224]   Address:    10.10.10.111:57168
[658704224]   Priority:   41819903
[658704224]   Foundation: 143138880
[658704224]   Username:   PppqqIrNy0J5ZkAq
[658704224]   Password:   ln38iPd7rG+WWbpaeJYiE9p7
```

**Step 4** `Janus` requests own `turnserver/PORT` from `turnserver` using `ICE` provided informations (username/password etc)

**Step 5** `Janus` push `RTP` packets to specific `turnserver/PORT` for negotiated `ICE` candidate using credentials `ice-ufrag/ice-pwd`.

**Step 6** `turnserver` relays received packet from `Janus` on negotiated ip/port e.g `10.10.10.10:45000` to the `Browser/ICE` on port e.g. `10.10.10.99:50000`

fyi @njam it must be still confirmed but looks like described

njam:
Maybe it has an effect only the other way around? When the browser send audio to the janus server?

kris-lab:
Maybe... I have tried to make it working with `audiobridge` but no luck so far. Looks like: if `janus` gets ICE from `turnserver` and sends to the browser then browser cannot push data to the `janus`. Other way around it works. Will investigate...

kris-lab:
I haven't manage to work Janus in `TURNSERVER` mode only (VM or any other unknown issue). **However it is not the scenario for us!**. Janus needs `TURNSERVER` only if placed behind the `NAT` (AWS etc.). As we gonna use public IP for  `Janus` then each client can push data directly to the `Janus` based on `SDP/ICE-candidates` offer from `Janus`. I would re-open this ticket regarding `SCALING` architecture. Maybe there is way to use TURN server for it and simply place `Janus` in private network only.

For scenario that Browser has no way for exposing IP or hole punching NAT on the network of any level then `TURNSERVER` works reliably. Browser get offer from `TURNSERVER` and sends to `Janus` which push data "locally" (via turnserver) instead of pushing directly to the client.

fyi @njam I am gonna consider the current state and knowledge as sufficient to proceed and close the topic for now in #1091

njam:
Makes sense, sounds good!

kris-lab:
closing in favour #1091

PR #1025, opened 2015-12-29
Janus: add recording/conversion tools

tomaszdurka:
The idea is to add new class which would install the script to convert `MJR` into useful `WebM` format. It requires 3 steps:
- convert `mjr/audio` -> `opus`
- convert `mjr/video` -> `webm`
- merge `opus/audio` and `webm/video` into new `webm/audio+video`

It requires `ffmpeg` module to be available + `janus` with enabled flag `--post-processing`

(copied from [janus](https://groups.google.com/forum/?pli=1#!topic/meetecho-janus/c8_JrDKu_w8))
```bash
bash
#!/bin/bash
#
# Copyright (c) 2015 Me
# Rev: PA1
# Convert local Janus .mjr files (audio and video) to an .webm file
# containing video and 2ch audio (vp8 + opus)
#
# dependencies:
#    janus-pp-rec (Janus postprocessing) must have been compiled and installed
#    ffmpeg (2015->) must have been installed.
#    (static versions for download exist on the internet)
#
# input is:
# mjr2webm -v myvideo.mjr -a myaudio.mjr out.webm
myhelp () {
    echo $0
    echo 'Usage: mjr2webm -v videofile -a audiofile [outputfile]'
    echo '   "videofile" must be a .mjr video file output from Janus'
    echo '   "audiofile" must be a .mjr audio file output from Janus'
    echo '   "outputfile", if not specified, is "default.webm"'
    echo '   if specified, it must end with filetype ".webm"'
}

video=-v
audio=-a
out=default.webm

if test $# -eq 0; then
   myhelp
   exit 0
fi

if ! which ffmpeg > /dev/null; then
   if ! which ./ffmpeg > /dev/null; then
     echo "Aborting: No 'ffmpeg' found" 1>&2
     exit 1
   else
      ffmpeg=./ffmpeg
   fi
else
   ffmpeg=ffmpeg
fi

if ! which janus-pp-rec > /dev/null; then
     echo "Aborting: No 'janus-pp-rec' found" 1>&2
     exit 1
fi

while test $# -gt 0
do
  case $1 in

  # Normal option processing
    -h | --help)
      # usage and help
      myhelp
      exit 0
      break
      ;;
    -v | --video)
      shift
      video=$1
      ;;
    -a | --audio)
      shift
      audio=$1
      ;;
  # ...
  # Done with options
    *)
      break
      ;;
  esac

  # for testing purposes:
  echo "param $1"

  shift
done


if ! [ -f $video ]; then
   echo "Videofile $video does not exist"
   exit 1
else
   if ! [[ "$video" == *.mjr ]]; then
      echo "Video file must end with .mjr"
      exit 1
   fi
fi
if ! [ -f $audio ]; then
   echo "Audio $audio does not exist"
   exit 1
else
   if ! [[ "$audio" == *.mjr ]]; then
      echo "Video file must end with .mjr"janus-pp-rec
      exit 1janus-pp-rec
   fi
fi
if [ -z "$1" ]; then
    echo "Using default.webm as output file"
else
    out=$1
    if ! [[ "$out" == *.webm ]]; then
       echo "Output file must end with .webm"
       exit 1
    fi
fi

# make temp files
TFILE="$$"
opus=$TFILE.opus
webm=$TFILE.webm

# convert audio to opus
janus-pp-rec $audio $opus

# convert video to webm
janus-pp-rec $video $webm

# mix the two:
# file gets large with high fr.
# by re-coding it we loose quality but reduce the bitrate to 1/4
# however, takes quite some time...
$ffmpeg -i $webm -i $opus -c:v libvpx -b:v 500k -c:a libopus -b:a 64k -r 25 $out

# fast modus
#$ffmpeg -i $webm -i $opus -c:v copy -c:a copy $out

# remove temp files
rm $opus
rm $webm
# done
echo "done!"
#------- end

```


tomaszdurka:
@kris-lab please review

tomaszdurka:
@kris-lab please review

kris-lab:
lvgtm!

- `libvpx -b:v 500k -c:a libopus -b:a 64k -r 25`: I would make it configurable with much higher default values e.g. `-b:v 5000k -c:a libopus -b:a 192k`. I think we would need to make it almost `lostless` on some point. Maybe we could optionally pass `ffmpeg` arguments as the string? `opus_ffmpeg_args=""` and `vp8_ffmpeg_args=""` @njam wdyt?

njam:
Yes, let's make it configurable! Agree on the proposed defaults.
Maybe also `set -e` in the script?

tomaszdurka:
@kris-lab pleae make final review

kris-lab:
@tomaszdurka quick note (note sure where to write it down). 

- `audioroom`: it dumps usable `WAV` format. There is nothing to do if we wanna keep it. Otherwise it needs to be converted to e.g. `MP3`
- `rtpbroadcast`: it dumps only `MJR` so current conversion to `WEBM` is required

So basically maybe we should add script for `WAV->MP3` ? cc @njam

kris-lab:
Missing `janus-pp-rec` in DEB package. Tracking https://github.com/cargomedia/debian-packages/issues/73

njam:
I would do the MP3 conversion, it will safe a lot of disk space. Or maybe opus? But probably mp3 is easier…

kris-lab:
- /usr/bin/mjr2png: line 48: ffmpegVideoParams: command not found: remove whitespace ` = ` helps
-  /usr/bin/mjr2png: `-w 768 -h 1024` is not working: `--height 1024 --width 768` works fine : problem is that `-h` is used for `help` and `height`. I would suggest to use `-hh` or `-x` (x-axis) for this?
- /usr/bin/mjr2webm: line 49: ffmpegVideoParams: command not found: remove whitespace ` = ` helps
- /usr/bin/mjr2webm: ffmpeg inputs have to be first so `ffmpeg -i audio.opus -i video.webm params-audio params-video output` instead of `ffpmeg -i audio.opus parms-audio -i video.webm params-video output` e.g. 
```
ffmpeg -i 27033.webm -i 27033.opus -c:v libvpx -b:v 500k -r 25 -c:a libopus -b:a 64k /vagrant/mjr_convert/video.webm`
```
- /usr/bin/mjr2webm: cannot convert only video like `mjr2webm -v /vagrant/mjr_convert/rec-1-3712273479-video.mjr /vagrant/mjr_convert/video-only.webm` -> `Audio file must end with .mjr``

After this small adjustments thee is proper output format for video/audio/png

kris-lab:
As @njam mention it would be good to add extra script for WAV->MP3 conversion! agree.

tomaszdurka:
@kris-lab made adjustments to your comments.

tomaszdurka:
So I should add something like:
```
ffmpeg -i input.wav -codec:a libmp3lame -qscale:a 2 output.mp3
```
into our module (mjr_converter - probably should rename it?).

tomaszdurka:
Or I should directly use ffmpeg in cm-janus?

njam:
Once question, in `mjr2webm.sh` we currently pass encoding parameters to ffmpeg (`-c:v libvpx -b:v 500k -r 25` and `-c:a libopus -b:a 64k`).
Maybe instead of decoding and re-encoding the video and audio, we could use `copy` to just use the original data?
http://superuser.com/questions/277642/how-to-merge-audio-and-video-file-in-ffmpeg/277667#277667

This could potentially be much faster, and lighter on the CPU, and we would have the same encoding like in the original streaming.

tomaszdurka:
Ok, I've done some testing/refactoring of mjr2webm script. It expects video and audio mjr sources which are parsed with `janus-rec-cc`. Then ffmpeg command copies streams chosing audio and video from each correct ones.

Example ffmpeg command:
```
ffmpeg -i 7257.webm -i 7257.opus -map 0:v -map 1:a -codec copy outmjr.webm
```

modules/ffmpeg/manifests/init.pp
```diff
@@ -1,4 +1,4 @@
-class ffmpeg ($version = '1.0') {
+class ffmpeg ($version = '2.6') {
```
njam: That's an upgrade :D

tomaszdurka:
@kris-lab please review

@njam @kris-lab I guess we should care about cpu usage of ffmpeg commands when running them in external apps.

kris-lab:
- mjr2png.sh: `--help` should be video not audio in the description?
- init.pp: maybe better use template("${module_name}/....") instead of `mjr_convert` 
- mjr2webm.sh: swap audio/video order hint

rest seem to work fine!

kris-lab:
Also what have we decided about WAV->MP3 ?

tomaszdurka:
@kris-lab adjusted to your comments. Let's merge this! :)

njam:
Let's set a (cpu) "nice" value!

tomaszdurka:
@njam or we could set it on cm-janus level, maybe more convenient.

njam:
sgtm

PR #1091, opened 2015-12-30
coturn continue of #952

kris-lab:
Part of #952 

Depends on https://github.com/cargomedia/puppet-packages/pull/1103


kris-lab:
@njam please have a look. At the end my changes are tiny. I have added credentials for user + more extended spec to make sure all is really applied.

As mentioned @ #952 all works as expected in case of authorisation, mobility, data flow from browser to `coturn` and back to the browser (back channel is our case a the moment).

njam:
- init.pp: $listening_ip: Shouldn't we default to empty array, which would listen on all addresses?
- init.pp: $relay_{min,max}_port: Rename to `$relay_port_{min,max}`?
- init.pp: $realm: Should this be optional? There's no `if` in the config, what will be the behaviour?
- config: `<% if @listening_ip …` not needed in addition to `each`
 - Same for `<% if @relay_ip`
 - Same for `<% if @external_ip`
- config: Don't we need to enable `lt-cred-mech` to use "long-term credential mechanism"? (see warning in log file)
- config: We should disable TLS/DTLS if we don't specify certificates (see warning in log file)
- init.pp: Increase number of open files (see warning in log file). Probably we should add it to `daemon`?
- init.pp: Remove `/etc/default/coturn` (probably used by an initv script, which we don't use?)
- init.pp: I would suggest to streamline configuration path to `/etc/coturn/turnserver,conf`
- init.pp: daemon should require config file
- init.pp: daemon: Inline $daemon_extra_args
- init.pp: Why is `$mice` disabled by default, is it a problematic feature?
- spec.rb: Add a check to make we have no warnings/errors in the log file?

kris-lab:
> $realm: Should this be optional? There's no if in the config, what will be the behaviour?

So far I am using it for setting user password.  e.g.
```
$ turnadmin --key --user njam --password njam --realm njam.com
0x009da43cb799f01739d6d3dabcee6bb6
```

`relam` is required for auth then! IMO we should keep it set on production

njam:
> relam is required for auth then! IMO we should keep it set on production

Ok, agree! Let's make it mandatory on the class then?

kris-lab:
> Don't we need to enable lt-cred-mech to use "long-term credential mechanism"? (see warning in log file)

hmm not sure. They say it should be default if any user is defined.
```
# This option is opposite to lt-cred-mech.
# (TURN Server with no-auth option allows anonymous access).
# If neither option is defined, and no users are defined,
# then no-auth is default. If at least one user is defined,
# in this file or in command line or in usersdb file, then
# lt-cred-mech is default.
#
#no-auth
```

njam:
If you check the log you'll see that they enable `lt-cred-mech` because there are users defined. Not sure anymore if it's a warning or not, just thought we should maybe make it explicit.

kris-lab:
@njam please have a look

njam:
- Use `ulimit::entry` (instead of class) - to avoid collisions
- Log still shows "Max number of open files/sockets allowed for this process: 4096"
 - Is it working? Should it rather be set in `daemon` (in the initv script, before spawning the process)?

Rest looks good

kris-lab:
> "Max number of open files/sockets allowed for this process: 4096"

Hmm the problem seems to be on the compilation step. see https://groups.google.com/d/msg/turn-server-project-rfc5766-turn-server/RHH3geNj-kI/mm0PiDJ43BoJ will verify...

kris-lab:
Confirmed. I could compile with `turnserver` with higher max for open files (`ulimit -n 65535`)

```
Version Coturn-4.5.0.3 'dan Eider'
0: 
Max number of open files/sockets allowed for this process: 65535
0: 
Due to the open files/sockets limitation,
max supported number of TURN Sessions possible is: 21500 (approximately)
0: 
```

kris-lab:
> "Max number of open files/sockets allowed for this process: 4096"

depends on https://github.com/cargomedia/debian-packages/issues/75

njam:
What really? 

> The default net engine is determined during the compile-time, not in the run-time

Usually the program should read the "nofiles" from the system, which is also what ultimately determines the limit?

ppp0:
interesting :)

kris-lab:
Really! :)

njam:
I think it's something with how we set the limit…
In a VM I can change it, and coturn picks it up:
```
# sudo -u turnserver sh -c 'ulimit -n 1234; turnserver -c /etc/coturn/turnserver.conf -v | grep "Max number of open files"'
[…]
Max number of open files/sockets allowed for this process: 1234
```

We should probably allow to pass "ulimit" options to `daemon`
- For systemd, see [docu](http://www.freedesktop.org/software/systemd/man/systemd.exec.html#LimitCPU=)
- For initv we can just execute `ulimit -something <value>` before running the daemon
(we're doing this for example in `pulsar_rest_api/templates/init.sh`)

kris-lab:
agree, I had the same result of my test.

kris-lab:
retest this please

kris-lab:
@njam please re-review

njam:
lgtm

PR #1092, opened 2015-12-31
Add bipbip logparser for cm-janus job failed

tomaszdurka:


tomaszdurka:
Multiline doesn't matter in the end, as we just want the occurrences of specific text-blocks.
@njam please review/reply to general error monitoring

njam:
- Let's define option `'metric_group' => 'cm-janus'` ?
 - Also add it for daemon? `'metric_group' => 'daemon'`
 - Check out the copperegg dashboard to see how it looks
- So "app error" will cover failing jobs?

tomaszdurka:
It will cover all `error` in general. We decided to use `error` only for those errors which should not occur. If they occur it's mostly cause connected services "forced invalid state".

tomaszdurka:
Or would you rather prefer to intercept invalid jobs errors separately?
I personally think having general error monitor is good. In perfect world we should never have those and if it happens it means there's "something" wrong and it requires investigation ;)

njam:
sgtm, let's adjust the metric_group though

tomaszdurka:
@njam please review

njam:
lgtm

tomaszdurka:
Looks good.

PR #1093, opened 2015-12-31
Fix log-parser for daemon

tomaszdurka:


tomaszdurka:
@njam please review. I presume this might fix invalid behaviour.

Another question should all these be in systemd, or they should be rather distributed into services?

njam:
lgtm, I think common group will make setting up alerts easier?

PR #1098, opened 2016-01-02
Add readline module

njam:


njam:
The studio-agent requires "readline", this was installed in my VM, because we add it to our images, [see here](https://github.com/cargomedia/vagrant-boxes/blob/2ec7a75898853354bfc6bdc2afb3a5597daba0f5/ubuntu-1504/scripts/base.sh#L6). But we don't have it on production.

njam:
@kris-lab please review

PR #1100, opened 2016-01-03
Fix "activesupport" version due to librarian-puppet dependency problem

njam:
@kris-lab @tomaszdurka please review


kris-lab:
hmm lgtm

PR #1101, opened 2016-01-03
Move rsyslog rules config to separate file, make sure we only log messages once

njam:
Currently rsyslog logs every message to /var/log/syslog twice, because such configuration is present in /etc/rsyslog.conf but also /etc/rsyslog.d/50-default.conf (ubuntu only).
As a resolution I'm moving our rules configuration to 50-default.conf always.

@kris-lab please review


kris-lab:
lgtm

PR #1102, opened 2016-01-04
Implement ufw class and utils

ppp0:


ppp0:
@njam please review

This speeds things up in that the first rule will match in the majority of cases (already established connections)

njam:
Nice btw: http://explainshell.com/explain?cmd=iptables+-A+INPUT+-m+state+--state+RELATED%2CESTABLISHED+-j+ACCEPT

- manifests/entry.pp: If we `include` only here, then the `exec` should require the package at least (because binary needs to be available). We can't use `require`?
- Let's lower case "icmp" in argument name?
- Should we have a new `class iptables::firewall` which has some convenience interface for us, to easily define rules like we do it in the softlayer interface, but in hiera? I imagine a list of hashes, each with:
 - source
 - destination
 - port (or port_min/port_max
 - protocol

ppp0:
* I got a circular dependency with `require` (because in `init.pp` an `entry` is now being created?)

njam:
Ok, then the entry should just require the binary/package?

ppp0:
@njam please review
@cargomedia/devops cc

njam:
- manifests/service.pp: This exec will probably be executed every time? Can you make sure no resources are applied on a second run?
- ufw/manifests/rule.pp: Ditto
 - Is this `::rule` really needed, or could we live with only `::application`? How do you plan to use these resources for implementing the firewall?

ppp0:
* I was thinking about how to prevent from running, but it would involve checking if the ufw is active, wich boils down to calling `ufw status` and then even grepping - which I thought to be more expensive than the idempotent activation?
* I was thinking to make this better - we definitely cound just define an application.d profile and have it activated implicitely

ppp0:
@njam please review

as for the repeated `exec` execution, I don't know, really. With `ufw allow X` it is the same problem (and it is idempotent, too)

njam:
We can't have resources that are applied repeatedly, it will trigger alerts.
I would go with `ufw status` then. The performance difference should be neglect-able.

njam:
You probably forgot to push?

ppp0:
Sorry, here the changes - I didn't think of the alert about repeated runs

@njam please review

njam:
lgtm

How do you plan to use these resources for implementing the firewall? Any additional wrapper needed?

ppp0:
good question.. some `create_resource` wrapping class would be needed

ppp0:
as discussed, let's use virtual resources and have them realized if  `ufw`  is included

PR #1109, opened 2016-01-07
Add optional firewalling to janus, coturn, cm-janus

ppp0:


ppp0:
https://github.com/cargomedia/puppet-packages/pull/1109

kris-lab:
- modules/daemon/Puppetfile: should go to `coturn`?
- modules/cm/spec/services/janus-firewalled-custom/manifest.pp: ssh rule is needed for specs?
- modules/cm/spec/services/janus-firewalled/manifest.pp: same like above?
- modules/cm/manifests/services/janus.pp what about plugin specific rules? I think `rtpbroadcast` opens RTP/UDP ports for listening -> puppet-packages/modules/janus/manifests/plugin/rtpbroadcast.pp?

kris-lab:
- modules/coturn/manifests/init.pp : what about `relay_port_min` and `relay_port_max`; I think there are defined for both `TCP/UDP` connection so should be opened as well?

ppp0:
* ups, deleted this
* port 22 is opened via custom rule
* added port 22
* plugin-specific rules.. Our needs are covered by this PR now, I can extend for completeness sake to plugins. Keep in mind, though, that we don't want to open any more ports as we're using them on the private network/localhost only
* modules/coturn/manifests/init.pp : this is not documented in the sk/readme - mistake?

ppp0:
So for `coturn` let's use the `application.d` profile that comes with the package
(https://github.com/cargomedia/debian-packages/blob/master/packages/vivid/coturn/debian/etc/ufw/applications.d/turnserver) - and use a custom rule only if provided

njam:
stopping old build, was running for 1h

ppp0:
hm, blocked port may be?

ppp0:
@devops I kinda like the default profile provided by the deb package for `coturn`. Shall we do the same for `janus` ? (In debian-packages?)

njam:
Aren't those ports configurable through puppet, so the profile needs to be updated too?

ppp0:
In case the ports are "customized" one would have to resort to pass the *complete* profile string to `ufw_app_profile`

njam:
Which is quite likely for janus, and also for coturn? I would configure in puppet therefore. Seems safer.

ppp0:
For `coturn` we thought to use the deb's default - but still be free to customize the profile (Is this needed for coturn? We weren't sure)

kris-lab:
hmm I would do `coturn` configuration based on puppet. Like I mentioned above
> modules/coturn/manifests/init.pp : what about relay_port_min and relay_port_max; I think there are defined for both TCP/UDP connection so should be opened as well?

njam:
@kris-lab could you take over here?

kris-lab:
closing in favour of #1115

PR #1115, opened 2016-01-08
(cont) firewall #1109

kris-lab:
cont of #1109


kris-lab:
@njam if you have moment please have a look. I think it is ok now...

njam:
- cm/manifests/services/janus.pp: Should we also define and pass along these two here?
 -   `$rtp_port_range_min = 20000,  $rtp_port_range_max = 25000,`
 - They are whitelisted on both UDP and TCP
- cm/spec/services/janus-firewalled-custom/manifest.pp: Don't use SSL, to simplify spec?
- coturn/manifests/init.pp: Shouldn't we also specify this profile directly here? I understand we're using a default config from the package now - but then it might not be in sync with the port configuration in puppet?

kris-lab:
> cm/manifests/services/janus.pp: Should we also define and pass along these two here?

in `cm::services`? or `janus` you mean? In `cm::services` we do not know this values...

njam:
If we add the firewall rules in `cm::janus` then we need to know all the ports there. IMO it's good to add `rtp_port_range_*` parameters to `cm::janus`, use it for the firewall, and pass it down to `janus`. wdyt?

kris-lab:
Hmm so we would add the Janus rule in cm::services::janus instead of Janus itself?

njam:
Ah no, like it is now is perfect, agree! Let's keep it in janus

- janus/manifests/init.pp: there's some double-$ in that ufw rule

kris-lab:
@njam please re-review

njam:
Sorry, I didn't look at this carefully. Isn't it strange that we define some of the rules in cm-janus, and some of them in janus? This way the rules in janus don't make sense themselves (because there's no websocket port available).
I think it would make sense to consider cm-janus to be the "public interface", and to define all ports and firewall rules there? Quick call?

kris-lab:
Well we define `cm-janus` specific port in `cm-janus` and `janus` specific ports in `janus`? let's have a call...

njam:
Yes if you use janus *inside* cm-janus it works, but if you would use only janus it wouldn't. So splitting the rules among the two modules can be confusing. I would suggest each module should have its own rules, but for janus we don't need the rules atm.

kris-lab:
ok, will re-do this by moving firewall stuff from janus to cm-janus only

kris-lab:
@njam please have a look

njam:
lgtm

njam:
(imo we could also leave this customization of the application profile away to simplify things, but up to you)

kris-lab:
will keep it for now!

PR #1117, opened 2016-01-08
Add metadata and puppetfile to Nvidia module

kris-lab:


kris-lab:
@njam please have a look

njam:
Should in require bipbip

kris-lab:
I was thinking as well:) Should we require virtual resources in puppetfile? I added this to be on safe site!

njam:
Yes it's needed, otherwise puppet will not know the resource.

kris-lab:
Funny, because I have applied this without `metdata` and `puppetfile` using `puppet agent -t` and it worked.

njam:
Yeah on production it usually works, because some other module already requires bipbip, so it's on the path?

kris-lab:
agree

PR #1119, opened 2016-01-11
UFW: manage /etc/ufw/applications.d ?

kris-lab:
Currently we have:
```
# ls -1 /etc/ufw/applications.d
cups
openssh-server
postfix
```

@kris-lab @ppp0 Which might not be what we want (e.g. postfix). Should we purge this directory with puppet, and add a rule to the ssh server module?

### DIFF
--- a/modules/ssh/Puppetfile
+++ b/modules/ssh/Puppetfile
@@ -5,3 +5,7 @@ mod 'cargomedia/helper',
 mod 'cargomedia/apt',
    :git => 'git@github.com:cargomedia/puppet-packages.git',
    :path => 'modules/apt'
+
+mod 'cargomedia/ufw',
+   :git => 'git@github.com:cargomedia/puppet-packages.git',
+   :path => 'modules/ufw'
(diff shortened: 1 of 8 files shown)


ppp0:
lgtm

kris-lab:
I have added `ufw::application::***` which would allow to add predefined services like `OpenSSH`.

@ppp0 @njam wdyt in general? 

@ppp0 please review (if you disagree, I will revert and we can merge only "purging")

ppp0:
Well, it looks good.. I don't have a strong opinion about the need to introduce a new class - I kinda fear it makes it less readeable than 
```
ufw::application { 'openSSH':
   app_or_port => '22',
}
```

but as I said, I'm neutral on this.. lgtm

kris-lab:
I am also ok with doing this in puppet-cargomedia. It will have to be duplicated for each server or group of server. With current approach we could keep it as `include`?

but as I said... both ways are ok with me

njam:
Why not put this into the SSH module, like we already do it for janus?

ppp0:
Agree, this is the best.. it's readable and it's tied to the original module. Drawback: whenever a `ufw` is included, the app profile gets activated

kris-lab:
Agree on adding this to the SSH.

@ppp0 please re-review. should be fine now.

ppp0:
Added test for firewall, renamed app Pprofile to what deb installation installs. @kris-lab please review

kris-lab:
lgtm

Issue #1071, opened 2015-12-14
Puppet 3.8 doesn't detect Service's "provider" correctly


njam:
For Ubuntu 15.04 detects either "sysvinit" or "upstart", but it should be "systemd".

Can be fixed by upgrading to Puppet 4.4: https://github.com/cargomedia/puppet-packages/pull/1068

Workarounds in place:
- puppet-packages: `site.pp` for "puppetmaster"
- sk: `bulldog.pp`

PR #1122, opened 2016-01-12
UFW - Fix order of exec for ssh and service recognition

ppp0:
* Fix Service provider (bug https://github.com/cargomedia/puppet-packages/issues/1071) - otherwise puppet won't be able to properly get `ufw`'s status
* Add before clause to https://github.com/cargomedia/puppet-packages/blob/master/modules/ssh/manifests/init.pp#L46 to make sure the rule is there *before* the firewall gets activated

### DIFF
--- a/modules/ssh/manifests/init.pp
+++ b/modules/ssh/manifests/init.pp
@@ -24,7 +24,7 @@ class ssh (
     notify  => Service['ssh'],
   }
 
-# Workaround for http://projects.puppetlabs.com/issues/2014
+  # Workaround for http://projects.puppetlabs.com/issues/2014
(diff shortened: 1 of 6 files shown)


ppp0:
@kris-lab please review

kris-lab:
Does it affect each "rule". Should it be done for each `ufw::application` and `ufw::rule`? I mean, something like global requirement for virtual resource? We do similar stuff in `apt` (`Exec['apt_update'] -> Package <| before != Exec['apt_update'] |>`)

ppp0:
good point!

ppp0:
@kris-lab please review

kris-lab:
that's lgtm

PR #1118, opened 2016-01-13
Daemon monitoring with systemd

ppp0:
@kris-lab @ppp0 @tomaszdurka 
Now that we run our daemons with systemd on Ubuntu we don't have any monitoring if all necessary daemons are running.
I think we should have some mechanism in place that reports either via email and/or CopperEgg.

I think it might be easiest to use monit after all. Maybe we could use monit to monitor, start and stop daemons, but leave the restarting to systemd.

wdyt?


ppp0:
* Under `systemd` monit now `pgrep`s with the binary's path - instead of pid for process check
* Start/stop are now `systemctl <start | stop>`
* Killing manually had `systemd` restart the process immediately. Since monit checks the process table, I'd say race conditions are minimized - Even if monit would be quicker to pick up that a process has to be restarted, it would invoke `systemctl start <process`> - looks safe to me
* Every process that we manage with the `daemon` class shall eventually have a dual (or systemd specific) monit template

@cargomedia/devops thoughts?

@kris-lab please review

ppp0:
Failing specs? On vm it looked good? On it

kris-lab:
tbh I have some concerns about `matching` with `monit`. Have you checked some production scenarios with `monit procmatch "regex-pattern"`?

> https://mmonit.com/monit/documentation/monit.html
CHECK PROCESS <unique name> <PIDFILE <path> | MATCHING <regex>>
<path> is the absolute path to the program's pid-file. A pid-file is a file, containing a Process's unique ID. If the pid-file does not exist or does not contain the PID number of a running process, Monit will call the entry's start method if defined.
<regex> is an alternative to using PID files and uses process name pattern matching to find the process to monitor. The first match is used so this form of check is most useful if the process name is unique. Pid-file should be used where possible as it defines expected pid exactly (pattern matching won't be useful for processes which start a child process using fork/clone as the child will match the same pattern temporarily). You can test if a process match a pattern from the command-line using monit procmatch "regex-pattern". This will lists all processes matching or not, the regex-pattern.

maybe it is safer to use `PIDFile` option for systemd? Then we are sure that we have correct PID?

ppp0:
You are right. An very good example is `cm-janus` (https://github.com/cargomedia/puppet-packages/blob/master/modules/cm_janus/manifests/init.pp#L73). The `regex-pattern` here, would be `/usr/bin/node`.

Issuing `sudo monit procmatch "/usr/bin/node"` on `pulsar sk production` yields
```
 ** [out :: stream2.fuckbook.com (10.55.40.158)] /usr/bin/node /usr/bin/socket-redis --log-dir=/var/log/socket-redis --redis-host=10.55.40.166 --socket-ports=8091,8092,8093,8094 --status-port=8085
 ** [out :: stream2.fuckbook.com (10.55.40.158)] /usr/bin/nodejs /usr/lib/node_modules/socket-redis/bin/socket-redis.js --socket-port=8091 --log-dir=/var/log/socket-redis
 ** [out :: stream2.fuckbook.com (10.55.40.158)] /usr/bin/nodejs /usr/lib/node_modules/socket-redis/bin/socket-redis.js --socket-port=8092 --log-dir=/var/log/socket-redis
 ** [out :: stream2.fuckbook.com (10.55.40.158)] /usr/bin/nodejs /usr/lib/node_modules/socket-redis/bin/socket-redis.js --socket-port=8093 --log-dir=/var/log/socket-redis
 ** [out :: stream2.fuckbook.com (10.55.40.158)] /usr/bin/nodejs /usr/lib/node_modules/socket-redis/bin/socket-redis.js --socket-port=8094 --log-dir=/var/log/socket-redis
 ** [out :: stream2.fuckbook.com (10.55.40.158)] ------------------------------------------
 ** [out :: stream2.fuckbook.com (10.55.40.158)] Total matches: 5
 ** [out :: stream2.fuckbook.com (10.55.40.158)] WARNING: multiple processes matched the pattern. The check is FIRST-MATCH based, please refine the pattern
 ** [out :: pulsar1.fuckbook.com (10.55.40.140)] List of processes matching pattern "/usr/bin/node":
 ** [out :: pulsar1.fuckbook.com (10.55.40.140)] ------------------------------------------
 ** [out :: pulsar1.fuckbook.com (10.55.40.140)] /usr/bin/node /usr/bin/pulsar-rest-api -c /etc/pulsar-rest-api/config.yml
 ** [out :: pulsar1.fuckbook.com (10.55.40.140)] /usr/bin/ssh-agent /usr/bin/node /usr/bin/pulsar-rest-api -c /etc/pulsar-rest-api/config.yml
 ** [out :: pulsar1.fuckbook.com (10.55.40.140)] ------------------------------------------
 ** [out :: pulsar1.fuckbook.com (10.55.40.140)] Total matches: 2
 ** [out :: pulsar1.fuckbook.com (10.55.40.140)] WARNING: multiple processes matched the pattern. The check is FIRST-MATCH based, please refine the pattern
 ** [out :: janus2.fuckbook.com (10.68.236.78)] List of processes matching pattern "/usr/bin/node":
 ** [out :: janus2.fuckbook.com (10.68.236.78)] ------------------------------------------
 ** [out :: janus2.fuckbook.com (10.68.236.78)] /usr/bin/node /usr/bin/cm-janus -c /etc/cm-janus/config.yaml
 ** [out :: janus2.fuckbook.com (10.68.236.78)] ------------------------------------------
 ** [out :: janus2.fuckbook.com (10.68.236.78)] Total matches: 1
cap> 
```

Even though, the match would be correct (there is only one node app running on janus2), this is too dangerous. How about adding the `args` (https://github.com/cargomedia/puppet-packages/blob/master/modules/cm_janus/manifests/init.pp#L74), too, if present?

Generating a PID is not supported by `systemd` - this needs to be done in a wrapping script. It is recommended only for `type=forking`. We use `type=simple` in our template. (cf http://www.dsm.fordham.edu/cgi-bin/man-cgi.pl?topic=systemd.service)

ppp0:
Excerpt from man page:
>    PIDFile=
           Takes an absolute file name pointing to the PID file of this
           daemon. Use of this option is recommended for services where Type=
           is set to forking. systemd will read the PID of the main process of
           the daemon after start-up of the service. systemd will not write to
           the file configured here, although it will remove the file after
           the service has shut down if it still exists.

ppp0:
The best would be if we could `check process with unit <service>`, there is an open issue (https://bitbucket.org/tildeslash/monit/issues/27/systemd-unit-support) about exactly this. Reading the comments in the discussions, it seems like the monit folks are up to do a major refactoring soonish anyway.

ppp0:
@kris-lab please review - I think it is now way safer. @njam please your opinion about this all

njam:
Hmhmhm I see, there's no PID file.
How about instead of checking for a process, we run a "check program 'systemctl status <name>'" - without start/stop?

njam:
Similar like planned in https://github.com/cargomedia/puppet-packages/issues/1065

ppp0:
Sound somewhat good to me - Restarting is handled by `systemd` anyway. What I would miss in such a scenario is the common interface `monit start` and `monit stop` but I could live with it

njam:
:+1: let's do it - monitoring is the important part imo

ppp0:
Another drawback is that deliberately stopping a process (such as we did with `monit stop process`) won't notify monit to stop monitoring

ppp0:
@kris-lab please review

njam:
> Another drawback is that deliberately stopping a process (such as we did with monit stop process) won't notify monit to stop monitoring

Can you even `monit stop foo`, if `foo` is a "check program"? Doesn't make sense I guess?

ppp0:
I reckon you can `monit unmonitor program`

ppp0:
Yeah, this is good news:

```
# monit stop my-program
# monit summary
The Monit daemon 5.11 uptime: 16h 8m 

Process 'postfix'                   Running
Program 'my-program'                Not monitored
System 'ubuntu-1504-amd64-default'  Running
```

So this mean, `monit stop <prgram instead of process>` is equivalent to `monit unmonitor` so that our pulsar recipes don't break! @cargomedia/devops cc

ppp0:
@kris-lab please review
@njam I'm not completely sure if it is a good idea as `monit stop` will just unmonitor but not stop a process under `systemd` leading to some rework of pulsar recipes

njam:
We don't use `monit stop` in any pulsar recipe (only monitor/unmonitor)?

kris-lab:
yes, agree, should be fine without reall `monit stop` support

njam:
- daemon-restart-on-kill: Don't restart twice on debian. Maybe check for syslog entry?

ppp0:
@njam: I find it difficult to verify that `monit` is reliably alerting. Every 10 seconds it's executing `systemctl status my-program` which would exit with non-null in the case of a *definitive failure only*. If the process is alive for a cpl of seconds only, chances are that monit won't pick it up but think all is good. I can `kill` the process and `sleep` but in the background, `systemd` will have restarted the process almost immediately and very likely not when `monit` will be polling the status....

njam:
That's fine, the main goal is to make sure a process gets stopped, and doesn't run at all and we don't notice it.

PR #1126, opened 2016-01-16
Open janus ports by default

njam:


njam:
@kris-lab @tomaszdurka please review

modules/janus/templates/config
```diff
@@ -1,5 +1,5 @@
 [general]
-interface = <%= @bind_address %>	; Interface to use (will be used in SDP)
+<% if !@bind_address %>;<% end %>interface = <%= @bind_address %>	; Interface to use (will be used in SDP)
```
tomaszdurka: Is this valid? I thought you are supposed to use unless :panda_face:

tomaszdurka: Or `if @bind_address.nil?` ?
I really thought this falsy value doesn't work reliably in ruby.

tomaszdurka:
Rest looks good.

njam:
retest this please

PR #1127, opened 2016-01-16
Make width/height parameters optional for mjr2png

tomaszdurka:


tomaszdurka:
@kris-lab please review

PR #1065, opened 2016-01-13
Reimplement RAID checks

ppp0:
- Don't use the wrapper scripts from "hwraid" anymore.
- Use the vendor binaries for each RAID controller, installable from https://github.com/cargomedia/debian-packages/:
 - arcconf
 - storcli (previously MegaCLI)
 - sas2ircu
 - hpacucli
- Check health with a simple `cli-command | grep "Optimal"` (or a wrapper perl, bash script), and run it regularly from monit:
```
CHECK PROGRAM bar PATH /bin/true
 IF STATUS != 0 THEN ALERT
```

cc @kris-lab @ppp0

PR #1130, opened 2016-01-18
[RAID monitoring] Add missing sas2ircu script

ppp0:
Follow-up of https://github.com/cargomedia/puppet-packages/pull/1065

### DIFF
--- a/modules/raid/manifests/lsi_megaraidsas.pp
+++ b/modules/raid/manifests/lsi_megaraidsas.pp
@@ -1,17 +1,24 @@
 class raid::lsi_megaraidsas {
 
+  require 'apt'
+  require 'apt::source::cargomedia'
   require 'python'
 
-  file { '/usr/local/sbin/lsi-raid-status':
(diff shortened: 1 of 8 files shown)


ppp0:
@njam please review

ppp0:
Another bug :/ - LSI monitoring does not work (affected: backup1)

ppp0:
Also check `sk-reflected` when done

njam:
- sas2ircu.pp: `File['/usr/local/sbin/sas2ircu-status']` probably requires the binary, so the package?
- specs are failing (exit code 3?)

ppp0:
* We're dumping the package from le vert - this is a python script? (https://github.com/ppp0/puppet-packages/blob/issue-1130/modules/raid/templates/sas2ircu/sas2ircu-status) and monit needs it to correctly check the status?
* Specs failing: on it

njam:
I meant "requiring" as int `requires => some other resource`.

PR #1136, opened 2016-01-20
Make temporary janus directories world-writable

njam:
@tomaszdurka please review

To delete a file you need write permissions on the containing directory (!).
I understand there are no sub-directories, but only flat files in these two folders, right?

cc @kris-lab


tomaszdurka:
Looks good, that's enough? :)

tomaszdurka:
666 :imp:

njam:
yes

PR #1131, opened 2016-01-20
[RAID] - Add hpssacli

ppp0:

PR #1137, opened 2016-01-21
Wrapping script for hpssacli

ppp0:
`hpssacli -exitonerror controller all show config` exits with 0 when the RAID is not good

Needs some more logic built into it

(Follow up of #1131)


ppp0:
@njam please check

njam:
Shouldn't grep be executed with `--invert-match`?

ppp0:
uh, of course!

ppp0:
@njam now?

njam:
lgtm

ppp0:
Uh oh! `--invert match` return 0 because it can match :(

PR #1139, opened 2016-01-21
Correct grepping of hpssacli status

ppp0:
Follow-up of #1137

### DIFF
--- a/modules/raid/templates/hpssacli/hpssacli-status
+++ b/modules/raid/templates/hpssacli/hpssacli-status
@@ -1,3 +1,7 @@
 #!/bin/bash -e
 
-hpssacli -exitonerror controller all show config | grep --invert-match -Ei 'failed|error'
+if (hpssacli -exitonerror controller all show config | grep -Ei 'failed|error'); then
+        exit 1;
+else
+        exit 0;
+fi


ppp0:
@njam please review

Tested on bulldog2 to work and to be green again when drive was reinserted, ignoring the rebuilding process, see:

```sh
  Port Name: 1I

   Port Name: 2I

   Internal Drive Cage at Port 1I, Box 6, OK

   Internal Drive Cage at Port 2I, Box 0, OK
   array A (Solid State SATA, Unused Space: 0  MB)


      logicaldrive 1 (111.8 GB, RAID 1, Recovering, 1% complete)

      physicaldrive 1I:6:1 (port 1I:box 6:bay 1, Solid State SATA, 120 GB, Rebuilding)
      physicaldrive 1I:6:2 (port 1I:box 6:bay 2, Solid State SATA, 120 GB, OK)
```

njam:
I don't understand, why is this change needed?

ppp0:
because `grep --invert-match` will match everything which is not `failed` or `error` and thus exit with 0..

njam:
Makes sense, lgtm!

PR #1140, opened 2016-01-22
Adjust cm-janus error log monitoring to new log4js logger

vogdb:


vogdb:
@njam please review

njam:
I understand a match looks like this:
```
2016-01-21 17:24:31.277 ERROR - <message>
```

The reason to match everything *before* "ERROR" is to not have false positives? In that case I think we should make sure that we match the beginning of the line as well?
So rather `'^[\d\-\:\s\.]+ERROR'`?

You can test it btw with:
```
irb
> '2016-01-21 17:24:31.277 ERROR - Error: Command failed:'.match(/^[\d\-\:\s\.]+ERROR/)
```

vogdb:
Good catch!

vogdb:
I tested on this
screen shot 2016-01-22 at 18 48 13

@njam please review

njam:
lgtm

PR #1141, opened 2016-01-22
Symlink ffmpeg to /usr/bin

njam:


njam:
@tomaszdurka fyi

cm-janus needs `ffmpeg` on the PATH. the PATH doesn't include `/usr/local/bin` in a daemon started by monit.
@ppp0 please remove this hack again when updating to DEB package.

ppp0:
as discussed, standard location when installed via deb package is `/usr/bin`.. see

```sh
vagrant@ubuntu-1504-amd64-default:~$ which ffmpeg
/usr/bin/ffmpeg
```

PR #1144, opened 2016-01-25
Add php SPL_Types extension

njam:
@fvovan wdyt?

### DIFF
--- a/modules/cm/manifests/application.pp
+++ b/modules/cm/manifests/application.pp
@@ -17,6 +17,7 @@ class cm::application (
   require 'php5::extension::mcrypt'
   require 'php5::extension::gearman'
   require 'php5::extension::mongo'
+  require 'php5::extension::spl_types'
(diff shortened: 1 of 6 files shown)


fvovan:
my only concern is 
>Warning
This extension is EXPERIMENTAL. The behaviour of this extension including the names of its functions and any other documentation surrounding this extension may change without notice in a future release of PHP. This extension should be used at your own risk.

http://php.net/manual/en/intro.spl-types.php

njam:
I would still give it a shot. Quite unlikely that they will change it  / drop it completely?
Otherwise we could do something similar in PHP (see https://gist.github.com/drock/6088355)?

fvovan:
sgtm

njam:
which one? :)

fvovan:
Ah. I mean standart for now :) and if they decide to deprecate it we will be able to switch

PR #1145, opened 2016-01-26
Daemon: make core dump configurable

kris-lab:


kris-lab:
@njam @ppp0 I am trying to make core dumps configurable on `daemon` level. The problem is with `start-stop-daemon` which seems to be not working with `PAM` (Pluggable Authentication Modules). Setting up limit with `/etc/security/limits.d/***` is not working properly.

I can hardcode `ulimit -c unlimited` into `init` script but this is not what we want?

any ideas?

njam:
What's the downside of adding `ulimit -c unlimited` to the init script? We do the same for nice, ulimit, oom_score.

kris-lab:
I thought it would be cleaner if `systemd` and `sysvinit` use `/etc/security/limits.d/***`. I cannot find any solution to apply `PAM` with `start-stop-daemon`. If you don't see other solution will go with hardcoding... it ok with me.

njam:
btw, maybe we should also set the "pattern", otherwise it will be default which is `core` - in the current directory, which is usually hard to find.

Also for systemd, should we the mechanism they provide? http://www.freedesktop.org/software/systemd/man/coredump.conf.html

kris-lab:
> "pattern"

I wanna do it in puppet-cargomedia (sysctl) similar to the spec https://github.com/cargomedia/puppet-packages/pull/1145/files#diff-d1693dc3f3e659931faca43eb43d0b68R5 (we set a lot of stuff in hiera, this one can be default as well)

> Also for systemd, should...

For `systemd` I think we can live with defaults, all is enabled anyway. It is just blocked by `ulimit`?

njam:
I understood systemd has its own way of storing core dumps (to its journal), which is not affected by ulimit settings. See for example: http://unix.stackexchange.com/questions/65110/no-more-coredumps-after-migrating-to-systemd

kris-lab:
True, I got confused by https://wiki.archlinux.org/index.php/Core_dump

Then all seems to be cleaner...

kris-lab:
quick update: not sure what I am doing wrong, but core dumping is not working for `systemd`. Someone has some experience, maybe wanna pair?

njam:
I have no experience unfortunately. The file is not created? Nothing in the journal?

Btw could we also check in `daemon-core-dump/spec.rb` that the dump was created?

kris-lab:
@ppp0 is up to help me. 

yes, I will check if dump is created (also for systemd) soon as will be there :) I cannot make it working :/

ppp0:
To me, the config looks good.. I don't get yet, why `systemd` does not dump to disk on `SIGSEGV`

ppp0:
I found the solution: http://serverfault.com/questions/582871/systemd-get-core-dumps-for-service-started-by-xinetd?answertab=active#tab-top and I begin to better understand the rants about `systemd`

ppp0:
Let's wait for CI, @kris-lab please have a look @njam cc

njam:
I think including `class { 'systemd::coredump': }` in `define daemon` will fail if we have multiple daemons? IMO we should include it once, in `systemd::init`.

kris-lab:
guys, it is not done at all. will mention if ready to review :)

njam:
sorry ;)

kris-lab:
@ppp0 please review

btw, I have removed `sysctl::entry { 'kernel-core-pattern':` from `systemd::coredump` + added to specs only. I think it is better to configure this globally in puppet-cargomedia [here](https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/files/hiera-data/defaults.json#L40) wdyt?

ppp0:
Have you tested what happens under `systemd` when `kernel-core-pattern` is not specified? I am not sure if it makes sense to enable coredumps but not to configure the sysctl entry (see https://sigquit.wordpress.com/2009/03/13/the-core-pattern/)? May be provide a default?

Rest looks good

kris-lab:
Well it is still working and dump core into `/core`. Agree it is not best solution but working properly. What would you suggest to do?

ppp0:
I would streamline it with wheezy, i.e. expect core dumps to be written to `/tmp` but this is just an opinion. @cargomedia/devops votes?

njam:
I would vote to default to  `/tmp/*` in puppet-packages, then we have it consistent across all our machines.

kris-lab:
Agree to have defaults. Some question:
- if we configure `kernel-pattern` in `systemd` module using `sysctl` then do you think puppet-cargomedia would be able to override this with [hiera](https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/files/hiera-data/defaults.json#L40) or would you set this value as `Systemd::Coredump { kore_dump_pattern => 'some-value' }`
- `sysvinit` how should we set defaults here?  By default is `core` so dumps to binary folder. If we hardcode `sysctl` to `init` script then we cannot easily override with global values.

wdyt?

njam:
Maybe this could be a class `sysctl::entry::core_pattern` with a sane default, and then be included from `daemon` (or `systemd` and `sysvinit`)?

kris-lab:
Makes sense to have sane default. Also to keep global configuration out of daemons. 

@ppp0 @njam please review

ppp0:
looks good to me. let's wait for ci to complete

PR #1146, opened 2016-01-27
Configure bundler install path

njam:
To avoid conflicts with binaries in the global bundler path
Related to: https://github.com/cargomedia/osx-setup/issues/105

@ppp0 @kris-lab please review
I suggest we stick to this convention in all repos where it makes sense

cc @tomaszdurka


ppp0:
lgtm

kris-lab:
totally agree!

tomaszdurka:
lgtm

PR #1148, opened 2016-01-27
Mongo: make auth configurable

kris-lab:
- [x] cluster auth support
- [x] basic auth support
- [x] mongorc support
- [x] mongorc::autologin support


kris-lab:
@njam please have a look on draft. The problem is that if we use `mongo` with `--eval` then `.mongorc.js` is not loaded. I had introduced hack but still testing. wdyt?

njam:
@tomaszdurka could you review this?

kris-lab:
@tomaszdurka it is not ready, need to be re-factored

kris-lab:
@tomaszdurka please review. I would suggest par-review for start. It is strongly related to https://github.com/cargomedia/puppet-cargomedia/pull/1044 and makes it hard to test.

kris-lab:
@tomaszdurka please hold on! I need to add spec for `cluster-auth` case.

kris-lab:
retest this please

modules/mongodb/lib/puppet/provider/mongodb.rb
```diff
@@ -41,7 +41,10 @@ def mongo_command(command, host, database = nil, options = {})
     defaults = {:skip_fail => false}
     options = defaults.merge(options)
 
-    args = ['--quiet', '--host', host, '--eval', command]
+    mongorc_path = options[:mongorc_path] || '/root/.mongorc.js'
```
tomaszdurka: If this is always run as root, I would not allow to change this `mongorc_path`.

modules/mongodb/manifests/core/mongod.pp
```diff
@@ -14,6 +16,20 @@
   $daemon = 'mongod'
   $instance_name = "${daemon}_${name}"
 
+  if $key_file_content {
+    $key_file_path = '/var/lib/mongodb/cluster-key-file'
+    if !defined(File[$key_file_path]) {
```
tomaszdurka: Not super happy about it, maybe a define with single argument would be bit better. Currently we check `filepath` only, but not content at all.

modules/mongodb/manifests/mongorc.pp
```diff
@@ -0,0 +1,13 @@
+define mongodb::mongorc (
```
tomaszdurka: Maybe this could be just a simple file?

modules/mongodb/manifests/mongorc/autologin.pp
```diff
@@ -0,0 +1,12 @@
+define mongodb::mongorc::autologin (
+  $database,
+  $username,
+  $password,
+  $system_user_homepath = '/root/.mongorc.js'
```
tomaszdurka: Maybe default to `/etc/mongo/mongorc.js` or whatever it was?

modules/mongodb/manifests/role/arbiter.pp
```diff
@@ -4,7 +4,8 @@
   $hostname = 'localhost',
   $repl_set,
   $repl_members,
-  $options = { }
+  $options = { },
+  $key_file_content = undef,
```
tomaszdurka: Maybe call it just `key`?

tomaszdurka:
Rest looks good. 

Not so happy about `getUsers` in mongorc auth script, but guess there's no better way.

modules/mongodb/manifests/role/arbiter.pp
kris-lab: Maybe something more specific like `cluster_key` or `auth_key`?

modules/mongodb/manifests/mongorc/autologin.pp
kris-lab: agree, can go to `/etc/mongorc.js`

modules/mongodb/manifests/mongorc.pp
kris-lab: Can be for now. Will require that each `mongorc` entry will have to contain `file` but it is fine.

modules/mongodb/manifests/core/mongod.pp
kris-lab: The reason for `!define()` is not the content of file but multiple declaration of `...core::mongod` if we apply multiple role per instance (like config1 which is config server and router). If content of file changes then puppet calculate md5 and applies changes.

modules/mongodb/lib/puppet/provider/mongodb.rb
kris-lab: Well it depends what would be the default in `mongodb::mongorc::autologin`. We have to keep it consistent so or `/etc/mongorc.js` or `/root/.mongorc.js`?

modules/mongodb/manifests/core/mongod.pp
kris-lab: Another option would be name "key-file" with prefix or suffix of instance name. It would create multiple-key files for multiple-role instances. It would solve the problem of multiple declaration of file resource (the same name as "path").

kris-lab:
retest this please

kris-lab:
@tomaszdurka let's pair on this? There are tiny topics to decide...

tomaszdurka:
Looks good to me.

PR #1151, opened 2016-02-04
Janus: core dumpe by default

kris-lab:


kris-lab:
@njam @ppp0 should we keep core dumps storage as default one `/tmp`? Or would you prefer to configure this in:
- `cm::services::janus` using `class { 'Sysctl::Entry::Core_pattern': pattern => '/tmp/foo.%e.%p.%h.%t' }`
- `sk::janus` (same config like above)
- `hiera` globally

ppp0:
I don't understand the question. Can't the default be overwritten? Every change is global to all processes, obviously

ppp0:
Change looks good, btw

PR #1159, opened 2016-02-08
Extend UFW/rule params with from/to

kris-lab:


kris-lab:
@ppp0 wdyt?

modules/ufw/manifests/application.pp
```diff
@@ -24,9 +25,10 @@
     refreshonly => true,
     notify      => Service['ufw'],
   }
-  ->
 
-  ufw::rule { "Allow ${app_name}":
-    app_or_port => $app_name,
+  if $auto_allow {
+    ufw::rule { "Allow ${app_name}":
+      app_or_port => $app_name
```
ppp0: `require => Exec["Refresh rule for ${app_name}"],`

modules/ufw/manifests/rule.pp
```diff
@@ -10,9 +13,20 @@
     default => 'allow',
   }
 
-  exec { "Set ${app_or_port} allow to ${allow}":
+  $proto_default = $proto ? {
```
ppp0: I think this is no `*_default` here, but a derived value which *can* be the default. How about calling the argument in line 6 `$tcp_or_udp` and use `$proto` instead of `$proto_default` ?

modules/ufw/manifests/rule.pp
```diff
@@ -10,9 +13,20 @@
...
+    undef   => '',
+    default => "proto ${proto}",
+  }
+
+  $is_port = ((count(split($app_or_port, ',')) > 1) or is_integer($app_or_port))
+  $app_or_port_default = $is_port ? {
```
ppp0: Similar issue here.. But more difficult naming.. How about leave the arg as it is and change from `$app_or_port_default` to `$target` ? I'm not sure.. but default isn't, agree?

modules/ufw/manifests/rule.pp
```diff
@@ -10,9 +13,20 @@
...
+    true => "port ${$app_or_port}",
+    default => "app ${app_or_port}",
+  }
+
+  exec { "Set ${app_or_port_default} allow to ${allow} from ${from} to ${to}":
     provider => shell,
-    command  => "ufw ${verb_allow} ${app_or_port}",
+    command  => "ufw ${verb_allow} ${proto_default} from ${from} to ${to} ${app_or_port_default}",
     unless   => "ufw status | grep -iqE '^${app_or_port}+.+${verb_allow}'",
```
ppp0: I think this `unless` needs to be reworked.. `$app_or_port` is probably not accurate and does it match correctly when  the same `app` is already defined elsewhere? I remember this was already a fragile heuristic when I first wrote it

ppp0:
* I think this is a viable approach to implement `to` and `from` restrictions.. See my comments above, though
* spec test for `auto_allow = false` ? To make sure it works and to have an usage example

ppp0:
I executed `bundle exec rake spec:ufw:app-profile os=Debian-7` to test..

If I dump `iptables -L`, the `ufw-user-input` looks like this:
```
Chain ufw-user-input (1 references)
target     prot opt source               destination         
ACCEPT     tcp  --  anywhere             anywhere             tcp dpt:ssh /* 'dapp_OpenSSH' */
ACCEPT     tcp  --  anywhere             anywhere             multiport dports ftp,telnet:smtp /* 'dapp_food' */
ACCEPT     udp  --  anywhere             anywhere             multiport dports 10000:15000 /* 'dapp_food' */
ACCEPT     udp  --  loopback/8           anywhere             udp dpt:25
ACCEPT     tcp  --  192.0.0.0/8          190.0.0.0/8          multiport dports telnet,24
ACCEPT     tcp  --  10.0.0.0/8           10.0.0.0/8           multiport dports ftp,telnet:smtp /* 'dapp_food' */
ACCEPT     udp  --  10.0.0.0/8           10.0.0.0/8           multiport dports 10000:15000 /* 'dapp_food' */
```

Afaiu, the top entries with `anywhere` will mask (ie make redundant) all following, because after the first match, no more evals will be done

This is the first and most important thing I notice. I'm reviewing the rest... Shall we pair on this quickly?

kris-lab:
@ppp0 please re-review

ppp0:
lgtm!

Also, by adding `auto_allow => false` to the spec manifest, we now get this order:
```
Chain ufw-user-input (1 references)
target     prot opt source               destination         
ACCEPT     tcp  --  10.0.0.0/8           10.0.0.0/8           multiport dports ftp,telnet:smtp /* 'dapp_food' */
ACCEPT     udp  --  10.0.0.0/8           10.0.0.0/8           multiport dports 10000:15000 /* 'dapp_food' */
ACCEPT     tcp  --  anywhere             anywhere             tcp dpt:ssh /* 'dapp_OpenSSH' */
ACCEPT     udp  --  loopback/8           anywhere             udp dpt:25
ACCEPT     tcp  --  192.0.0.0/8          190.0.0.0/8          multiport dports telnet,24
ACCEPT     tcp  --  anywhere             anywhere             multiport dports re-mail-ck:60
ACCEPT     tcp  --  anywhere             anywhere             multiport dports 100:150,200
```

makes more sense! Still we can't protect the user from entering garbage!

kris-lab:
@ppp0 there were some bugs. Please re-check.

ppp0:
uh! of course!

lgtm

PR #1161, opened 2016-02-10
[nginx] - Make server_names_hash_bucket_size configurable

ppp0:
On a virtual node it seems nginx has problems detecting the proper CPU cache line size (see https://trac.nginx.org/nginx/ticket/352)

If virtual hostnames become too long, you'll get the error
```
[emerg] 13253#0: could not build the server_names_hash, you should increase server_names_hash_bucket_size: 32
```

(This on a virtual node that has `getconf LEVEL1_DCACHE_LINESIZE == 64`)


ppp0:
@njam please review

njam:
- What's the reasoning for  1daf6b5? `hasrestart` does *not* default to `true`. I think `enable` might also be needed?

rest lgtm

ppp0:
OK, reverted

njam:
Where was it reverted?

ppp0:
Oh, I did not push :blush:

PR #1164, opened 2016-02-11
[jessie] Look into php extensions that fail

ppp0:
* As of now, on Wheezy, the official deb package bears version `Version: 3.1.0~rc1-1+b2`
*For some reason, we download and compile version 3.1.2 from pecl (see https://github.com/cargomedia/puppet-packages/blob/master/modules/php5/manifests/extension/imagick.pp#L2)
* Official Jessie deb package version is `3.2.0~rc1-1`
* Current stable version is `3.3.0` (see https://pecl.php.net/package/imagick)

Keeping the same setup (manually build 3.1.2) brought me some problems because of the build-depend `libmagickwand` whose jessie version breaks the build. We could **patch the configure script to point to the correct location** of `MagickWand-config` which before was to be found in `/usr/bin` - see https://github.com/rflynn/imgmin/issues/51 and `/usr/share/doc/imagemagick/NEWS.Debian.gz`:
```
imagemagick (8:6.8.9.9-1) unstable; urgency=high

  Obsolete config scripts (Magick-config, MagickCore-config,
  MagickWand-config, Wand-config and Magick++-config) are
  not multi-arch safe and thus have been removed from /usr/bin.
  .
  Moreover, these scripts are superseded by pkg-config
  facilities.
  .
  However as a courtesy to our users, these scripts have
  been moved to
  /usr/lib/$DEB_HOST_MULTIARCH/ImageMagick-$VERSION/bin-$QUANTUMDEPTH/
  where $DEB_HOST_MULTIARCH is the multi-arch triplet and
  $QUANTUMDEPTH is the current quantum depth. $VERSION is the upstream
  version without the revision number.
  .
  Please note that these scripts will be definitively removed after
  jessie.

 -- Bastien Roucariès <roucaries.bastien+debian@gmail.com>  Sun, 19 Oct 2014 12:12:20 +0200
```

I would install 3.2.0-RC1 as per Jessie repository and test if we need to upgrade to `3.3.0`

@cargomedia/devops comments welcome


ppp0:
@njam please review

njam:
Install config in both cases, so we can add configuration if needed?

ppp0:
So I reintroduced this config and things started to irk
Will fix next week, this whole php5 config thing is flawed

ppp0:
`php5::extension::apc` not needed anymore after #1166

ppp0:
php5::extension::opcache not needed in jessie (built-in in php 5.6)

```
php -v
PHP 5.6.17-0+deb8u1 (cli) (built: Jan 13 2016 09:10:12) 
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2015, by Zend Technologies
```

ppp0:
some changes here:
* apc was only needed by revive (depends on #1166)
* opcache is a built-in in php for jessie (v5.6)
* opcache configurabiity
* bipbip plugin is named `apc` therefore `php5::fpm::with_apc` not renamed (still needed for apcu)

@njam please review

njam:
- extension/imagick.pp: What is the reason to have this `ini_content` param?
- extension/opcache.pp: Looks a bit strange that we guard the installation with `lsbmajdistrelease`. I think in general we should rather use `lsbdistcodename` (or `lsbdistcodename` *and* `lsbmajdistrelease`) for it to be unique. For this specific case maybe it's best to add a fact that returns the PHP version?
- php5/manifests/fpm.pp: What remove `config_extension_change` - the purpose is to restart php-fpm when the configuration for an extension changes. I think we should still have it?

ppp0:
* I thought it would make sense to be able to modify imagick's ini file - but it doesn't, it's enough to manage it
* sgmt
* In my understanding, this was some sort of debug class unless there's more magic behind it:
```puppet
class php5::config_extension_change {

  exec { 'echo Triggering refresh because php extensions configuration changed':
    provider    => shell,
    refreshonly => true,
  }
}
```

njam:
`php5::config_extension_change` - the purpose is to restart php-fpm when the configuration for an extension changes! (via `notify` and `subscribe`)

ppp0:
uh oh, I tried to introduce a `php_version` fact which would return a hash such as 
```rb
{"release" =>"5.6", "full"=>"PHP 5.6.17-0+deb8u1 (cli) (built: Jan 13 2016 09:10:12)"}
```

The problem is that before php is installed there is no way to get its version by issuing `php- v`

ppp0:
Alright, ci passes, changes applied

@njam please review

njam:
- manifests/extension/imagick.pp: The extension should be installed before the config (to make sure we only restart php-fpm once the extensions is available)
- extension/opcache.pp: use "codename"
- manifests/fpm.pp: Keep subscribing to "change"
- php5/manifests/init.pp: Use "codename"
- php5/manifests/init.pp: Why include opcache here?

ppp0:
@njam another review, please

njam:
lgtm

But depends on  https://github.com/cargomedia/puppet-packages/pull/1166

PR #1166, opened 2016-02-16
Migrate revive to nginx

ppp0:
To drop apache2 as a dependency, let's do the move.

see
https://gist.github.com/metafour/a69d7c67724a0523d6f9
https://www.digitalocean.com/community/tutorials/how-to-install-openx-on-centos-6


ppp0:
next step: migrate to jessie

ppp0:
phase out `php5::extension::apc` (in favour of apcu and opcache)

ppp0:
@njam please review

ppp0:
This will need some luv when merging. Likely, apache needs to be stopped and uninstalled first

njam:
- manifests/init.pp: This `expires 1y` looks weird - this will apply to any request I guess? We didn't have it in apache, and I think it's something the application (revive) should decide about for reach request?
- manifests/init.pp: Where is this gzip mimetype list coming from? Is it good to zip PNGs?

ppp0:
* Removed
* the list's original is here: https://github.com/cargomedia/puppet-packages/blob/master/modules/cm/manifests/vhost.pp#L113 - I thought it to be good but it can be removed, also, a `gzip_min_length` could be defined. I'd say gzipping statics is good practice?

ppp0:
@njam please another look

njam:
Images like PNG are usually already compressed, so I think it's not common to zip them again (little to no gain). I would suggest to stick to the original list, as here: https://github.com/cargomedia/puppet-packages/blob/master/modules/cm/manifests/vhost.pp#L113

Otherwise lgtm

ppp0:
retest this please

PR #1172, opened 2016-02-17
Gearmand with daemon instead of init scripts

ppp0:


ppp0:
@njam please review (after ci pass)

njam:
CI fails, but otherwise lgtm

PR #1167, opened 2016-02-11
Migrate gearman service to daemon

ppp0:
needed by #1150

PR #1169, opened 2016-02-18
Drop unused /etc/aliases to manage email forwards in postfix

ppp0:
Via https://github.com/cargomedia/puppet-packages/pull/1167
E.g. with:
```
bundle exec rake spec:cron os=Debian-7 keep_box=true
```

Before:
```
vagrant@debian-7-amd64-default:~$ cat /etc/aliases
postmaster: root
monit: root
```
After:
```
vagrant@debian-7-amd64-default:~$ cat /etc/aliases
postmaster: root
```

### DIFF
--- a/modules/postfix/manifests/init.pp
+++ b/modules/postfix/manifests/init.pp
@@ -59,23 +59,6 @@ class postfix ($aliases = { }, $transports = []) {
     notify      => Service['postfix'],
   }
 
-  file { '/etc/aliases':
-    ensure => file,
-    content => template("${module_name}/aliases"),
-    group  => '0',
-    owner  => '0',
-    mode   => '0644',
-    notify  => Exec['newaliases'],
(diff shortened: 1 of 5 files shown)


ppp0:
@njam please review

njam:
lgtm

njam:
Is there any mail sent to "postmaster"? If so we should probably add this forward?

ppp0:
Seem these entries need to be there in order to comply with RFC 5321 (see https://tools.ietf.org/html/rfc2142#section-5). I'm not sure shall we put them statically into the template (as per this PR), configure it in https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/files/hiera-data/defaults.json#L25 or another option

@njam wdyt?

njam:
lgtm as you did it

PR #1150, opened 2016-02-04
Add Jessie to missing modules for ci deployment

ppp0:
Still to be done: - for https://github.com/cargomedia/puppet-cargomedia/issues/1034

- [x] copperegg_revealcloud
- [x] network
- [x] augeas
- [x] sysctl
- [x] iptables

Other forgotten (and needed) modules:
- [x] vagrant (solved in #1153)
- [x] virtualbox
- [x] raid
- [x] gearman (~~#1167~~ #1172)

PR #1165, opened 2016-02-18
[Jessie] - Migration of composer and php5

ppp0:
Follow-up of https://github.com/cargomedia/puppet-packages/pull/1150
Needs #1164 https://github.com/cargomedia/puppet-packages/issues/1168

Make following modules run on Jessie:
- [x] composer
- [x] php5


ppp0:
@njam please review

we're not using suhosin, are we?

njam:
Agree! Then we can probably also drop it for wheezy and remove it completely?

ppp0:
@njam  please review

njam:
lgtm

PR #1175, opened 2016-02-19
Revert "Move mongorc/autologin to /root/.mongorc.js by default"

kris-lab:
Reverts cargomedia/puppet-packages#1173


kris-lab:
Too many issues in `pulsar` behaind this small change! Will try one more time after all.

PR #1181, opened 2016-02-29
Add "janus instance" define

kris-lab:
Via https://github.com/cargomedia/puppet-cargomedia/issues/1080

Depends on https://github.com/cargomedia/puppet-packages/issues/1196 https://github.com/cargomedia/puppet-packages/issues/1197

For scaling janus we'd like to run multiple janus-gateway servers on the same machine. So we need to refactor the "class" to a "define".

Idea:
- Introduce "janus instance" define
 - Make all paths configurable (log, etc, var - or maybe just prefix?)
 - Make global resources use a unique *instance name* (daemon entry, bipbip entry, …)
- Keep the "class", but make it use the "janus instance"

Then eventually we can install multiple *instances*, e.g. to `/opt/janus-cluster/instance{1,2,3}`.

Idea:
```puppet
define janus::instance (
  $title = "foo",
  $log_file = '/janus3/var/log/janus/janus.log',
  $config_file = '/janus3/etc/janus/janus.cfg',
  $rtp_port_range_min = 10000,
  $rtp_port_range_max = 11000,
)
```

@kris-lab @ppp0 wdyt?

---

[kris-lab notes]

TODO:
- [x] plugin/transport libraries are stored in common location. Every `janus` process loads plugins and look for configuration. It is going to fail! We need way to configure plugins and transport per process/instance.


kris-lab:
@ppp0 I have done some progress in this ticket. Would you be up for pairing on this and then we decide who is gonna finish it up?

kris-lab:
TODO:
- [ ] implements symlinks for plugins and transports
- [ ] update specs 
- [ ] make sure janus.cfg in latest version
- [ ] implement ROLES

ppp0:
@kris-lab I got it to work (roughly), will continue tomorrow

ppp0:
@kris-lab please take over

ppp0:
Reconfiguring of `cm::service::janus` from here https://github.com/cargomedia/puppet-packages/blob/master/modules/cm/manifests/services/janus.pp#L27

kris-lab:
@njam wanna have a look

njam:
cm_janus:
- cm_janus/manifests/proxy.pp: Doesn't the `proxy_pass` value need to match the upstream definition (`cm::upstream::proxy` a bit up)?
 - I would not use `cm::upstream::proxy` at all here, but just plain `nginx::resource::upstream`.
- cm_janus/spec/init/1-install.pp: Regarding `removeAllStreams()`, let's link to https://github.com/cargomedia/cm-janus/issues/201
- cm_janus/spec/: How about adding a spec that installs two instances? This would help to avoid resource conflicts, which can be difficult to catch otherwise.

janus:
- janus/manifests/core/janus.pp: How about renaming `janus::core::janus` to `janus::server` (janus/server/init.pp) and `janus::core::setup_dirs` to `janus::server::setup_dirs` (janus/server/setup_dirs.pp)?
- janus/manifests/core/setup_dirs.pp: There's an `->` between `Exec` and `File` *and* there's a `require => Exec` on every file. I guess one should be sufficient?
- janus/manifests/core/setup_dirs.pp: `$ssl_config_dir` is defined but not used (rather the value is calculated again)
- janus/manifests/core/setup_dirs.pp: Should we recursively purge this directory? We're managing it completely… 
- janus/manifests/plugin/audioroom.pp: Isn't `Class['janus::common'] -> Janus::Plugin::Audioroom[$name]` the same as `require 'janus::common`?
 - Same for rtpbroadcast and transports
- janus/spec/cluster/manifest.pp: Inline those variables at the beginning? Would make it more readable imo.
- janus/spec/cluster/manifest.pp: Could we simplify the configuration of $prefix by specifying it only once (e.g. on `Janus::Core::Janus`), and the other resources would read it there?
- janus/spec/from-src/spec.rb: todo
- janus/templates/config: Remove the disabling of default plugins, now that we have the symlinking mechanism?
- cm/manifests/services/janus.pp: I understand the goal is that we can have two of these installed on the same node. Then we need to make sure that sub-resources have unique names (cm_janus, cm_janus::proxy). Maybe add a spec?
- cm/spec/services/janus-default/spec.rb: todo

kris-lab:
>  janus/manifests/core/setup_dirs.pp: Should we recursively purge this directory? We're managing it completely…

We have logs there as well?

njam:
true

kris-lab:
I have stacked here https://github.com/kris-lab/puppet-packages/blob/issue-1181/modules/cm/spec/services/janus-cluster/manifest.pp#L77

We need `cm-api` for running `cm-janus`. It is done with `nginx` but unfortunately `cm:serivces:janus` uses `nginx` as well with `cm-janus:proxy`. It interferes and results with not starting first of `standalone` janus as `nginx` restarts in meantime :/

njam:
You mean for the `removeAllStreams()` call?
It will be resolved and released in https://github.com/cargomedia/cm-janus/pull/201 asap.

kris-lab:
yes! I cannot wait so I have found this `nginx` mock solution. However with multiple cm-janus, janus instances it is not working anymore.

let's wait then!

njam:
- [ ] janus/manifests/core/setup_dirs.pp: Why have defaults? These params are always passed.
 - What about those variables `_default` - they are never defined?
- [ ] janus/manifests/core/janus.pp: How about renaming `janus::core::janus` to `janus::server` (janus/server/init.pp) and `janus::core::setup_dirs` to `janus::server::setup_dirs` (janus/server/setup_dirs.pp)?
- [ ] janus/manifests/core/setup_dirs.pp: There's an `->` between `Exec` and `File` *and* there's a `require => Exec` on every file. I guess one should be sufficient?
- [ ] janus/manifests/plugin/audioroom.pp: Isn't `Class['janus::common'] -> Janus::Plugin::Audioroom[$name]` the same as `require 'janus::common`?
 - [ ] Same for rtpbroadcast and transports
- [ ] janus/spec/cluster/manifest.pp: Inline those variables at the beginning? Would make it more readable imo.
- [ ] janus/spec/cluster/manifest.pp: Could we simplify the configuration of $prefix by specifying it only once (e.g. on `Janus::Core::Janus`), and the other resources would read it there?
- [ ] janus/spec/from-src/spec.rb: todo
- [ ] janus/templates/config: Remove the disabling of default plugins, now that we have the symlinking mechanism?

kris-lab:
@njam please have a look.

kris-lab:
It is failing due to `apt-get` problems.

retest this please

PR #1190, opened 2016-02-29
Backup copy network interface - To be able to revert

ppp0:
In case of mistakes in network config we would be able to revert by manually restoring the backup copy of `/etc/network/interfaces`

### DIFF
--- a/modules/network/manifests/init.pp
+++ b/modules/network/manifests/init.pp
@@ -1,5 +1,13 @@
 class network {
 
+  exec { 'Backup existing network config':
+    provider => shell,
+    command  => 'cp /etc/network/interfaces /etc/network/interfaces.provisioned',
+    unless   => 'ls /etc/network/interfaces.provisioned',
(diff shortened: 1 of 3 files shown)


ppp0:
@kris-lab please review
@cargomedia/devops cc

kris-lab:
agree, it is safe to have this feature.

lgtm

ppp0:
@njam I let have you the last word ;)

njam:
- Should make sure that this runs *before* we modify the original file
- The name is a bit unclear to me. Maybe just "interfaces.bak"?

ppp0:
@njam please review

njam:
Shouldn't this be one resource to require at the beginning of `network::interface` (instead of requiring it on every change)? Maybe a new `network::interfaces_backup`?

ppp0:
agree much more elegant now. @njam please review

njam:
lgtm

PR #1191, opened 2016-03-01
Allow to set cm-janus version

tomaszdurka:


tomaszdurka:
@ppp0 please revuew

ppp0:
lgtm

PR #1200, opened 2016-03-09
Fix jobs_path for plugins

kris-lab:


kris-lab:
paired with @tomaszdurka

PR #1209, opened 2016-03-14
cm::services::janus: Expect boolean value for recording_enabled

tomaszdurka:


tomaszdurka:
@njam please review

njam:
Closes https://github.com/cargomedia/puppet-packages/issues/1210

tomaszdurka:
Removed your comment about closing #1210. Guess mistype.

tomaszdurka:
Closes #1211

njam:
How about also using boolean for the plugins' signatures? I think it will be easier if we have the same everywhere.

tomaszdurka:
I've came back to separating issues.

@njam please review

njam:
lgtm

PR #1210, opened 2016-03-14
janus: Create jobs_path directory

tomaszdurka:


tomaszdurka:
@kris-lab or @njam please review

modules/cm/manifests/services/janus.pp
```diff
@@ -59,6 +59,7 @@
     plugin_rtpb_minport              => $rtpbroadcast_minport,
     plugin_rtpb_maxport              => $rtpbroadcast_maxport,
   }
+  ->
```
tomaszdurka: It's actually all about:

```puppet
File [$jobs_path] -> Daemon['cm-janus_video']
```
but I think we should not go into specifics.
I could also add `require => File [$jobs_path]` (only if this file defined...).

tomaszdurka:
Actually moved the ordering into cm-janus.

njam:
- What if "plugin_jobs_path" is `undef` in standalone?
- Via https://github.com/cargomedia/sk/issues/3070: btw instead of `/var/lib/janus/jobs`, shouldn't it be instance-specific? So for the video installation and for the audio installation?

tomaszdurka:
@njam Please review all together, I didn't want to revert things so ended up all together.

njam:
- file { $plugin_jobs_path: Why group "0" (and not "janus")?
- What if "plugin_jobs_path" is `undef` in standalone?
- Via https://github.com/cargomedia/sk/issues/3070: btw instead of `/var/lib/janus/jobs`, shouldn't it be instance-specific? So for the video installation and for the audio installation?

tomaszdurka:
- Ok
- Well if it's single instance it will be ok, if you have multiple you need to adjust it as in sepc.
- It is now per instance. One for all the plugins from single instance.

tomaszdurka:
@njam please look

njam:
As discussed, should use the directory managed in `janus::server::setup_dirs`?

tomaszdurka:
Solved in #1214

PR #1204, opened 2016-03-14
Add janus-gateway-cluster streaming roles

kris-lab:
Part of #1181 and https://github.com/cargomedia/puppet-cargomedia/issues/1080


kris-lab:
@tomaszdurka have a look on this as well. This is part of https://github.com/cargomedia/janus-testing/pull/9

For now I call it `cluster_member` for simplicity and testing but we need really decide on `origin`, `repeater`, `edge` role which in case of `origin` needs also `cm-janus`, `cm-janus-proxy` etc...

tomaszdurka:
- Still not sure why we have `cluster_member` and not directly `origin`, etc.
- Wdyt about separating cm-janus from origin and rather have job-manager only therE?

kris-lab:
This is half way point. We need rather origin, repeater, edge roles but all of them have similar interface (params) as cluster_member currently. I have tried to use puppet "inherits" to get list of params to "define" but puppet does not supports this!

Separation of job mamager from cm-janus is a great idea. However, does it make sense to have origin role without cm-janus and janus-cluster-manager? All this roles without orchestration mechanism have no meaning really?

modules/janus/templates/role/streaming/register
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+curl -XPOST -H 'Content-Type: application/json' -d '<%= @content %>' <%= @url %>
```
tomaszdurka: I guess this one is not used, but rather https://github.com/kris-lab/puppet-packages/blob/issue-1204/modules/janus_cluster_manager/templates/register_member.sh.erb.

tomaszdurka:
Looks ok, shall we add some specs for `role::rtpbroadcast`?

kris-lab:
@tomaszdurka please have a look one more time - all note very welcome!

kris-lab:
Merging. I need this for doing faster progress...

PR #1212, opened 2016-03-14
Only run apt-update for packages by *apt*

njam:
@ppp0 please review


ppp0:
makes sense and looks good

PR #1213, opened 2016-03-14
Require "apt" for "percona_toolkit" because installer uses DPKG

njam:
dpkg used in the installer, so we need to make sure apt indices are up-to-date.

@ppp0 please review


ppp0:
Good point. Same situation for modules:
* mms
* elasticsearch
* vagrant

?.. more?

njam:
@ppp0 please review

ppp0:
lgtm

PR #1188, opened 2016-03-14
Jenkins wants to run tests for deleted folders

ppp0:
To reproduce:

- remove a folder which had specs before
- PR

test run will fail (spec not found)

https://github.com/cargomedia/puppet-packages/blob/master/build-ci.sh#L18 should not run if there is a folder deletion

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -67,8 +67,9 @@ namespace :spec do
     end
     module_list.compact!
     module_list.uniq!
+    module_list.select! { |module_name| finder.puppet_module?(module_name) }
     specs = module_list.map do |module_name|
-      finder.puppet_module(module_name).specs if finder.puppet_module(module_name)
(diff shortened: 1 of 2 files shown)


ppp0:
@njam please review

njam:
I would rather keep the raising behaviour of `puppet_module`, so that other usages fail early. Also I'm not sure if this even works, because `puppet_module` still returns a non-false value?

How about introducing `Finder#puppet_module?(name)` which return boolean, and then using it to filter the module list?
```ruby
module_list.select! { |module_name| finder.puppet_module?(module_name) }
```

ppp0:
Ok then tomorrow

ppp0:
Makes sense, @njam please review

njam:
please adjust formatting in Rakefile

otherwise good to go

njam:
lgtm

PR #1215, opened 2016-03-16
Introduce rsyslog entry

ppp0:


ppp0:
needed by #1198 

@njam please review

njam:
I would suggest to call it `rsyslog::config` for consistency?
Otherwise lgtm

PR #1220, opened 2016-03-23
Adjust cm-janus fluentd log tailing

njam:


njam:
@kris-lab some fixes for fluentd for cm-janus.
Merging to fix things.
Please revie!

PR #1203, opened 2016-03-24
Janus: Load modules before daemon starts

kris-lab:
Seems that modules are loaded after initial start of janus.

Currently we have this:
```puppet
  file { "${instance_base_dir}/usr/lib/janus/transports.enabled/libjanus_websockets.so":
    ensure    => link,
    target    => '/usr/lib/janus/transports/libjanus_websockets.so',
  }

  file { "${instance_base_dir}/etc/janus/janus.transport.websockets.cfg":
    ensure    => 'present',
    content   => template("${module_name}/transport/websockets.cfg"),
    owner     => '0',
    group     => '0',
    mode      => '0644',
    before    => Daemon[$instance_name],
    notify    => Service[$instance_name],
  }
```

I guess we should ensure module is loaded before Daemon.

@kris-lab agree?

### DIFF
--- a/modules/janus/manifests/transport/websockets.pp
+++ b/modules/janus/manifests/transport/websockets.pp
@@ -31,6 +31,7 @@ define janus::transport::websockets(
     ensure    => link,
     target    => '/usr/lib/janus/transports/libjanus_websockets.so',
   }
+  ->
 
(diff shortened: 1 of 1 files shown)


kris-lab:
@njam please review. There is missing ordering arrow. For the rest of plugins and transports it is correct.

njam:
lgtm

PR #1219, opened 2016-03-24
cm-janus daemons should subscribe to package

kris-lab:
So it restarts on upgrades.

@kris-lab or @tomaszdurka can you look into this?


kris-lab:
not sure what is the problem exactly but if restarts on upgrade then should work...

@njam please have a look

njam:
lgtm

PR #1221, opened 2016-03-24
Migrate satis to nginx (from apache2)

ppp0:
Parf of https://github.com/cargomedia/puppet-cargomedia/issues/1114

PR #1223, opened 2016-03-29
Disable standard page for apache2/puppetmaster

ppp0:
Follow-up of https://github.com/cargomedia/puppet-packages/pull/1221

On the same box, there is apache running as front-end to puppetmaster. Implement redirector http->https in apache?

[UPDATE]

apache2/puppetmaster only needs 8141, the other default ports are live but returning the default "It works" page.. Renaming this issue

### DIFF
--- a/modules/apache2/manifests/init.pp
+++ b/modules/apache2/manifests/init.pp
@@ -32,4 +32,13 @@ class apache2 {
     ensure   => present,
     provider => 'apt',
   }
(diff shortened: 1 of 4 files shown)


ppp0:
@njam please review

ppp0:
satis and puppetmaster are currently running due to a manual intervention

njam:
lgtm

PR #1224, opened 2016-03-30
Add fluentd loggly plugin

njam:
@vogdb wanna review?

Can be used like this:
```
  fluentd::config::source{ 'my-source-1':
    type     => 'tail',
    config   => {
      path   => '/tmp/foo',
      format => 'json',
      tag    => 'foo',
    },
  }

  fluentd::config::match{ 'foo':
    type     => 'loggly',
    pattern  => 'foo.**',
    config   => {
      loggly_url => 'https://logs-01.loggly.com/inputs/<token>/tag/foo',
    },
  }
```


vogdb:
ok

PR #1225, opened 2016-03-31
Refactor logrotate::entry

ppp0:
Refactor modules:
- [x] fluentd
- [x] pulsar-rest-api
- [x] bipbip
- [x] cm-janus
- [x] coturn
- [x] janus
- [x] mongodb
- [x] php5
- [x] socket-redis
- [x] ufw
- [x] mysql


ppp0:
@njam please review

njam:
- Since `rotation_newfile => 'copytruncate'` is so popular, how about making it default?
- How about introducing a param `postrotate`, so we don't need `additional_config`?
- Let's streamline rotation for all services. I think differences are mostly random. Let's do for example 14 days and daily everywhere?
- php: `$logrotate_postrotate` is not used

ppp0:
retest this please

ppp0:
The only spec failing is
`Error: /Stage[main]/Main/Node[default]/Mongodb::Core::Mongos[router1]/Exec[wait for mongos_router1 up]: Command exceeded timeout`

IMO this can be tolerated

@njam please review (I'll pull upstream, so the CI will run again for hours)

njam:
- Does `logrotate::entry` really require the log file's directory to be present? What happens if it's not there yet?
- logrotate/spec/entry/manifest.pp: formatting
- mongodb: Does not use `copytruncate`, but instead has this `kill -USR1`. Now with the refactoring it's also using `copytruncate` (because it's default). Is that on purpose? Or should we set the previous default for mongodb explicitely?
- Can you look into the failing mongos spec? Would be good to fix it / make it more robust.

ppp0:
* It fails silently with exit 0. Iv'e removed the require
* ok
* I changed it to the former default 'create' (but I don't think of this as a disrupting change)
* I streamlined the timeout to 300 secs for both `mongos` and `mongod`. Let's hope it's enough

ppp0:
@njam please review (long-running CI)

njam:
- socket_redis/manifests/init.pp: Revert `->` before `logrotate::entry`?

Rest lgtm

ppp0:
hm, the specs were failing on mongo, again, this time with a deb signature key download problem?
```
Notice: /Stage[main]/Mongodb/Apt::Source[mongodb]/Apt::Key[mongodb]/Exec[Add deb signature key for mongodb]/returns: gpg: requesting key 7F0CEB10 from hkp server keyserver.ubuntu.com
Notice: /Stage[main]/Mongodb/Apt::Source[mongodb]/Apt::Key[mongodb]/Exec[Add deb signature key for mongodb]/returns: gpgkeys: key 7F0CEB10 not found on keyserver
Notice: /Stage[main]/Mongodb/Apt::Source[mongodb]/Apt::Key[mongodb]/Exec[Add deb signature key for mongodb]/returns: gpg: no valid OpenPGP data found.
Notice: /Stage[main]/Mongodb/Apt::Source[mongodb]/Apt::Key[mongodb]/Exec[Add deb signature key for mongodb]/returns: gpg: Total number processed: 0
Error: apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 7F0CEB10 returned 2 instead of one of [0]
Error: /Stage[main]/Mongodb/Apt::Source[mongodb]/Apt::Key[mongodb]/Exec[Add deb signature key for mongodb]/returns: change from notrun to 0 failed: apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 7F0CEB10 returned 2 instead of one of [0]
```

PR #1226, opened 2016-04-04
Daemon: Allow other signals than SIGTERM

ppp0:
Make stop signal configurable

### DIFF
--- a/modules/daemon/manifests/init.pp
+++ b/modules/daemon/manifests/init.pp
@@ -8,6 +8,7 @@ define daemon (
   $env = {},
   $limit_nofile = undef,
   $core_dump = false,
+  $forced_exit = false,
 ) {
 
   if (defined(User[$user])) {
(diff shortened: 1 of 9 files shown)


ppp0:
@njam please review

njam:
- systemd.service.erb: Why change [`Restart=`](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=) to "no"?

ppp0:
To reflect the `force_stop` in a systemd environment (and thus make the spec test succeed)

njam:
- Rename "forced_exit" to "sysvinit_kill"
- New test with TERM-ignoring program. Check for both cases (sysvinit and systemd) that process doesn't run anymore.

ppp0:
@njam please review

njam:
lgtm, could you enable it for fluentd?

PR #1229, opened 2016-04-04
Fail in MJR scripts if no output file was generated

njam:


njam:
Paired with @tomaszdurka

njam:
@kris-lab fyi

PR #1231, opened 2016-04-04
Add fluentd source for syslog

njam:


njam:
@ppp0 please review

njam:
The format comes from http://docs.fluentd.org/articles/in_tail - I just adjusted the time format, because we use "high precision timestamps"

ppp0:
lgtm

PR #1236, opened 2016-04-07
Fluentd: map different levels to a default

njam:
To unify the different levels

http://tools.ietf.org/html/rfc5424
https://github.com/Seldaek/monolog/blob/master/src/Monolog/Logger.php
http://ruby-doc.org/stdlib-2.1.0/libdoc/logger/rdoc/Logger.html
https://github.com/cargomedia/cm/blob/master/library/CM/Log/Logger.php


njam:
Paired with @tomaszdurka

PR #1222, opened 2016-04-08
Reconsider cm-janus logparser

njam:
It's the third time that the pattern is not correct anymore :|
```
    - name: error
      regexp: "^[\x5cd\x5c-\x5c:\x5cs\x5c.]+ERROR"
```

Should we rather remove it, and do this kind of stuff through fluentd?


njam:
@tomaszdurka please review

tomaszdurka:
lgtm

PR #1247, opened 2016-04-11
Purge unmanaged xsession definitions

njam:
Addressing https://github.com/cargomedia/puppet-cargomedia/issues/1148

This will remove any unmanaged xsession configurations.
So the unity entry will be removed from the greeter, to avoid confusion.
I think we don't need unity. If we need it, we can start it manually with `gnome-session --session=ubuntu`.

@kris-lab wdyt?


kris-lab:
I think we don not need `unity` but let's double check!

njam:
@kris-lab I've added a "unity" module, then we could install that on staging so we have the ability to still use it.

wdyt?

kris-lab:
I think it is good!

Issue #1242, opened 2016-04-06
Make modules Debian 8 compatible


ppp0:
Part of https://github.com/cargomedia/roadmap/issues/95

- [x] apt
- [x] autoprefixer
- [x] bash
- [x] bipbip
- [x] browserify
- [x] ca_certificates
- [x] database
- [x] ffmpeg
- [x] janus
- [x] lame
- [x] mysqltuner
- [x] nodejs
- [x] systemd
- [x] helper
- [x] build
- [x] cgi_fcgi
- [x] cm (#1254)
- [x] cm_janus
- [x] composer
- [x] coturn (#1384)
- [x] cron
- [x] daemon
- [x] elasticsearch
- [x] environment
- [x] fluentd
- [x] foreman
- [x] gearman
- [x] git
- [x] github
- [x] iptables
- [x] irqbalance
- [x] janus
- [x] kernel
- [x] lame
- [x] logrotate
- [x] memcached
- [x] mjr_convert
- [x] mongodb (https://github.com/cargomedia/puppet-packages/pull/1371)
- [x] monit
- [x] mysql (#1243)
- [x] mysql_proxy
- [x] mysqltuner
- [x] network
- [x] nginx
- [x] ntp
- [x] percona_toolkit
- [x] phantomjs
- [x] php5
- [x] phpunit
- [x] postfix
- [x] puppet
- [x] raid
- [x] redis
- [x] rsync
- [x] rsyslog
- [x] ruby
- [x] socket_redis
- [x] ssh
- [x] sudo
- [x] sysctl
- [x] timezone
- [x] ufw
- [x] uglify
- [x] ulimit
- [x] vim

ppp0:
depends on https://github.com/cargomedia/puppet-packages/issues/1243

ppp0:
depends on https://github.com/cargomedia/puppet-packages/issues/1244

njam:
I've added a list to the description with puppet modules that are currently installed in SK, but only available for Debian 7.

kris-lab:
All modules seems to be compatible with Jessie.

(coping some notes from [Janus ticket](https://github.com/cargomedia/puppet-packages/issues/1346#issuecomment-238521428)
- [ ] disable mongodb logs #1430
- [ ]  disable logs for janus #1377
- [ ]  foreman-debian cargomedia/pulsar-conf-cargomedia#331

kris-lab:
[...]

PR #1244, opened 2016-04-12
Port janus and plugin modules to jessie

ppp0:


ppp0:
@njam please review

ppp0:
part of #1242

njam:
Why change those specs? The old version covers more ground I think.
@kris-lab can you review this, as you're the original author?

kris-lab:
@ppp I guess the only change is that you check all ports together for listening (not per janus process)?

ppp0:
I did not change the specs just for fun. Under `systemd` - with our `daemon`-generated service description files - there is no such thing as a `PID` file, so the spec would break. `netstat` on the contrary, knows which port is listening and what process is bound to it. I cheerfully welcome better solutions

ppp0:
spec fails because of https://github.com/cargomedia/puppet-packages/pull/1249 which is way more important than this PR here - @kris-lab please review

kris-lab:
retest this please

ppp0:
@kris-lab please review

kris-lab:
lgtm

njam:
Thanks for the explanation. A short note either as a comment or a commit message would be helpful, so the reviewer knows the reasoning if it's not obvious from the code.

Now we could probably also just use [serverspec's matcher](http://serverspec.org/resource_types.html#port):
```ruby
describe port(80) do
  it { should be_listening.with('tcp') }
end
```

ppp0:
I thought about the matcher port(N) but since you can't give it a list of ports to check it would be too verbose for my taste

PR #1251, opened 2016-04-14
Introduce execpre and post to daemon

ppp0:


ppp0:
@njam please review

njam:
As *daemon* is an abstraction module it would be good to keep the public interface and the specs as generic as possible.
- Pre/post: we can easily run this also on sysvinit, just put it into the script. For the spec we can check if those tmp files are created (instead of checking `systemctl status`)
 - To simplify the spec how about inlining those pre and post commands?
- PermissionsStartOnly: I suggest we run pre/post commands always as root, and just hardcode this to `true` for systemd?

njam:
Two commands, that will be executed as root for both systemd and sysvinit:
```pp
  $pre_command = undef,
  $post_command = undef,
```

(`PermissionsStartOnly = true` always)

ppp0:
@njam please review

njam:
- daemon/spec.rb: I would drop this systemd-specific check. It's not needed (we cover the functionality on a higher level with `my-program-*` and it's only for systemd?
- templates/sysvinit.sh.erb: Let's put those commands in `if` so we don't have empty lines if not set. Also helps clarifying to developers that these are optional.

ppp0:
@njam please review

njam:
lgtm

PR #1252, opened 2016-04-15
Add postgresql module

njam:
Needed for https://github.com/cargomedia/puppet-packages/pull/1068

Inspired by https://github.com/puppetlabs/puppetlabs-postgresql

### DIFF
--- a/modules/postgresql/spec/server-database/spec.rb
+++ b/modules/postgresql/spec/server-database/spec.rb
@@ -1,0 +1,10 @@
+require 'spec_helper'
+
+describe 'postgresql::server::database' do
+
+  describe command('sudo -u postgres psql -t -c "SELECT datname FROM pg_database;"') do
+    its(:stdout) { should match 'foo' }
+    its(:stdout) { should match 'bar' }
+  end
+
+end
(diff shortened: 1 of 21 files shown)


njam:
@ppp0 @kris-lab Adding a basic postgresql module (server and client), with defines for "role" , "database" and "extension". This will be needed for the new puppetdb for puppetserver.
Please review.

modules/postgresql/spec/server-database/spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+describe 'postgresql::server::database' do
+
+  describe command('sudo -u postgres psql -t -c "SELECT datname FROM pg_database;"') do
+    its(:stdout) { should match 'foo' }
+    its(:stdout) { should match 'bar' }
+  end
+
```
ppp0: May also check for role `alice`, since you create it in the manifest?
see http://www.postgresql.org/docs/9.4/static/database-roles.html

njam: This test is about the `database` resource, there's a separate one for the `role`. I'm just creating a role here, because one cannot create a database without a role. So it's more like setting up prerequisites, but not something we should assert on?

ppp0: true.

ppp0:
lgtm

PR #1250, opened 2016-04-18
Don't install "latest" for gst plugins

kris-lab:
@kris-lab @ppp0 currently we install the "latest" gst plugin DEBs.

Should we change that, to have finer control over upgrades?

### DIFF
--- a/modules/gstreamer/manifests/plugins/bad.pp
+++ b/modules/gstreamer/manifests/plugins/bad.pp
@@ -1,12 +1,9 @@
-class gstreamer::plugins::bad (
-  $version = 'present',
-) {
+class gstreamer::plugins::bad {
 
   require 'apt::source::cargomedia'
   require 'gstreamer::plugins::base'
 
   package { 'gstreamer1.0-plugins-bad':
-    ensure => $version,
     provider => 'apt',
   }
 }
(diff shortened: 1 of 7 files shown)


kris-lab:
@njam please review

ppp0:
looks a bit weird to have a "present" version.. No way to default to `undef` and translate this later when `package { 'foo': ensure => present }` is called?

kris-lab:
you can use `ensure` with value `present` or `installed` https://docs.puppet.com/puppet/3.8/reference/type.html#package-attribute-ensure

njam:
imo we could just remove this "version" param, because the only values that can reasonably be passed are "present" and "latest" (but not any version string, and not "absend"). So I'd say it's rather confusing to have it in the first place.
wdyt?

ppp0:
OK for me to completely remove it

kris-lab:
@njam @ppp0 what about now?

njam:
We usually specify `ensure => present`, but I guess it's optional.

retest this please

ppp0:
it's not optional according to `lint`? Iirc we need to specify it.. or was it `provider => 'apt'`...

ppp0:
retest this please

PR #1256, opened 2016-04-18
Port redis-server to jessie

ppp0:


ppp0:
@njam please review

njam:
- manifests/init.pp: Let's name the configs according to the redis version. `redis-2.4.conf`, `redis-2.6.conf`?
- manifests/init.pp: Where are these sysctl values coming from? I can find references for all kinds of values, for both versions.
- manifests/init.pp: Bipbip entry should require Daemon instead of Package
- spec/default/spec.rb: Let's remove the monit check, this is covered in daemon's specs
- templates/monit.debian.erb: unused
- templates/monit.systemd.erb: unused

ppp0:
* good idea!
* excerpt from redis-2.6.conf: 
   >
```
# TCP listen() backlog.
#
# In high requests-per-second environments you need an high backlog in order
# to avoid slow clients connections issues. Note that the Linux kernel
# will silently truncate it to the value of /proc/sys/net/core/somaxconn so
# make sure to raise both the value of somaxconn and tcp_max_syn_backlog
# in order to get the desired effect.
tcp-backlog 511
```
* ok
* not fully agree: I wanted to check that the different templates wheezy vs not behave well
* see below
* not fully agree: implicitely called by `daemon` module

ppp0:
@njam please review

njam:
lgtm

I didn't test the no-daemonized handling of redis-server, but I trust you checked it?

ppp0:
monit needs a config refresh:
* /etc/init.d/monit stop
* rm /etc/monit/conf.d/redis
* /etc/init.d/monit start
* redis-cli shutdown (to shut down old process)

ppp0:
done on 
* www3
* denkmal

PR #1261, opened 2016-04-20
Upgrade Node.js to v5.x

njam:
From 0.12

Used by:
- autoprefixer
- bower
- browserify
- cm-janus
- less
- pulsar-rest-api
- socket-redis
- uglify

@vogdb fyi

### DIFF
--- a/modules/nodejs/manifests/init.pp
+++ b/modules/nodejs/manifests/init.pp
@@ -4,8 +4,8 @@ class nodejs {
 
   apt::source { 'nodesource':
     entries => [
-      "deb https://deb.nodesource.com/node_0.12 ${::lsbdistcodename} main",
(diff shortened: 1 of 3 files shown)


njam:
@vogdb please review

I ran the specs of all affected modules (see above). I checked cm-janus specifically. inode is working, it picks up new job definitions as they appear on the filesystem.

njam:
@ppp0 or you want to review?
btw also making Node.js available for Debian 8 here.

vogdb:
Why did you put here the upgrade of debian to 8?

njam:
Nope, just making the Node.js module compatible with Debian 8.

@vogdb @ppp0 good to go?

vogdb:
lgtm from me

PR #1266, opened 2016-04-21
Run "vagrant plugin install" in home dir

njam:
@ppp0 please review

The process writes/reads to/from CWD it seems, so fails if it's /root.


ppp0:
uh oh , lgtm

PR #1269, opened 2016-04-22
Add shell script to shut down and delete all virtualbox VMs

njam:
@kris-lab please review
@ppp0 fyi

When we stop builds manually sometimes we end up with "zombie" VMs on CI.
The only reliable way to stop and delete them is to use `VBoxManage`. `vagrant global-status` does not necessarily see them.
This script should shut down and delete all VMs.
Use carefully.


kris-lab:
I have made test with skipping `--delete`!

lgtm

PR #1270, opened 2016-04-23
Fixes for puppet4

njam:
Cherry picked from https://github.com/cargomedia/puppet-packages/pull/1068


njam:
Requires puppet4, will be done in https://github.com/cargomedia/puppet-packages/pull/1068

PR #1274, opened 2016-04-29
Make accountsservice spec reliable

njam:
@kris-lab please review

The trick seems to be that one has to *request/find* a user first, before setting any configuration on it


kris-lab:
lgtm

PR #1272, opened 2016-04-29
Add Xorg module

kris-lab:
For https://github.com/cargomedia/puppet-packages/issues/973

Allow to add entries to /etc/X11/xorg.conf:
```
        ModulePath   "/usr/lib/nvidia-346-updates/xorg/"
        ModulePath   "/usr/lib/nvidia-352/xorg/"
```

Have a look at other puppet modules?

Should be required by "lightdm".

---
@kris-lab notes
- https://github.com/joshbeard/puppet-display
- http://augeas.net/docs/references/lenses/files/xorg-aug.html
- http://augeas.net/docs/references/1.4.0/lenses/files/xorg-aug.html#Xorg.Configuration_files
- https://gist.github.com/ex0hunt/c433d590b72961706fe7
- https://docs.puppet.com/puppet/3.8/reference/type.html#augeas
- https://projects.puppetlabs.com/projects/1/wiki/puppet_augeas


kris-lab:
@njam please have a look

It works reliable if we ADD/SET new elements in the section. It check if elements exist and skips INSERT for true. 

Currently it doesn't update value for any key of any section. wdyt?

njam:
Cool! I think that's what we need.

- xorg::common: Just call it "xorg"?
- xorg::common: Install package "xorg"
- xorg::config: Require instead of include augeas, because it's required bit down anyway
- xorg::config: Require instead of include xorg, because it's needed for the config to be in place
- xorg::config: Augeas define: Include "xorg::config" in name, to ensure uniqueness
- spec/config/file/manifest.pp: Remove "file" from path
- spec/config/file/manifest.pp: Those specs suggest we want to use a "sub" config file (in "conf.d") for configuring multitouch, right?
 - xorg::config: If so, then it would be nice if we don't need to pass in the full config path, but just the name of the subconfig

kris-lab:
> spec/config/file/manifest.pp: Those specs suggest we want to use a "sub" config file (in "conf.d") for configuring multitouch, right?

Currently this config goes to  `/usr/share/X11/xorg.conf.d/50-multitouch.conf`. But agree it should be doable to move into `/etc/X11/xorg.conf.d/50-multitouch.conf` and simplify interface of `xorg::config` resource.

should we do it?

njam:
sgtm, so `xorg::config` takes an optional parameter `$config_name = undef`?

kris-lab:
@njam wdyt now? (I messed up my commit/naming, sorry)

njam:
Maybe squash commits? :)

- manifests/config.pp: Rename $xorg_path_final to $config_path?

rest lgtm

kris-lab:
ok, makes sense!

njam:
We should also notify the xorg server when the config changes!
Would also be nice to check that the configuration is actually applied in the spec. Maybe by checking that some text shows up in xorg's log?

Could you also verify that it works correctly together with lightdm - if we restart the xorg server, probably lightdm needs to restart as well? does it happen automatically?

kris-lab:
> We should also notify the xorg server when the config changes!

agree

> Would also be nice to check that the configuration is actually applied in the spec. Maybe by checking that some text shows up in xorg's log?

this sounds like testing if X server actually consider configuration etc. do you really wanna test this basics?

> Could you also verify that it works correctly together with lightdm - if we restart the xorg server, probably lightdm needs to restart as well? does it happen automatically?

hmm will verify but imo the `lightdm` starts X so restarting `lightdm` or eg. `gdm` (for gnome) will restart `X` with new configuration.

PR #1278, opened 2016-04-30
Remove "tagmail" report (not supported anymore)

njam:


njam:
@cargomedia/devops fyi

There's no more "tagmail" (email) report in puppet. Fortunately we have reporting+alerting with bipbip.

ppp0:
@njam I notice you do the same here in the `exec` as I'm doing in https://github.com/cargomedia/puppet-packages/pull/1671

Is/Was there a reason to use `exec` instead of managing the directory?

modules/puppet/manifests/agent.pp
```diff
@@ -25,6 +25,13 @@
       notify  => Exec['/etc/puppetlabs/puppet/puppet.conf'];
   }
 
+  exec { 'chmod 0755 /opt/puppetlabs/puppet/cache':
```
ppp0: @njam sorry, this is the code I was meaning in https://github.com/cargomedia/puppet-packages/pull/1278#issuecomment-281123958

njam: hm sorry, don't remember - and commit message doesn't explain it

PR #1295, opened 2016-05-02
Make screenconnect installation check more robust

njam:


njam:
@ppp0 please review

`dpkg -l <package>` will exit `0` even if the package is *uninstalled* (but was installed before), so now I'm checking if the "status" is "installed".

ppp0:
Makes sense! Lgtm

PR #1297, opened 2016-05-02
Add phantomjs module

ppp0:
To run QUnit tests (https://github.com/cargomedia/cm/pull/2140), we need to have [phamtomjs][phantomjs] installed in dev and CI env.

Some inspiration: https://forge.puppet.com/3fs/phantomjs

[phantomjs]: http://phantomjs.org

### DIFF
--- a/modules/phantomjs/Puppetfile
+++ b/modules/phantomjs/Puppetfile
@@ -1,11 +1,6 @@
+mod 'puppetlabs/stdlib',
+  :git => 'git://github.com/puppetlabs/puppetlabs-stdlib.git'
+
 mod 'cargomedia/helper',
   :git => 'git@github.com:cargomedia/puppet-packages.git',
   :path => 'modules/helper
-
-mod 'cargomedia/apt',
-  :git => 'git@github.com:cargomedia/puppet-packages.git',
-  :path => 'modules/apt'
-
-mod 'cargomedia/build',
-  :git => 'git@github.com:cargomedia/puppet-packages.git',
-  :path => 'modules/build'
--- a/modules/phantomjs/Puppetfile
+++ b/modules/phantomjs/Puppetfile
@@ -1,0 +1,11 @@
+mod 'cargomedia/helper',
+  :git => 'git@github.com:cargomedia/puppet-packages.git',
+  :path => 'modules/helper
+
+mod 'cargomedia/apt',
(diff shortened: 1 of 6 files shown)


ppp0:
Here our installer (just `require 'phantomjs'`)

This installs a statically-linked binary. According to http://phantomjs.org/download.html#linux-64-bit I've also required to package `fontconfig` but without being able to verify its correct installation / need

@njam please review

njam:
- Why is "build" needed?
- Defining the "unp" package will conflict with puppet-cargomedia. How about using original tooling to do the job?
 - Alternatively we could use this stdlib function [ensure_package](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/ensure_packages.rb) they use in the linked module.

ppp0:
* oops.. you're right - a leftover from some gist I took an initial inspiration of
* agree `tar xjf` does the job
* I put it on, what are the upsides exactly versus using `package` which does an existence check, too?

@njam please re-review

njam:
Having two `package{'foo':}` in the catalog will result in a conflict. This function seems to check if the same resource is already defined (with the same parameters).

One more thing: Will `install.sh` work when an (older) version is already installed? Maybe just `rm -rf /usr/local/share/phantomjs` first?

njam:
Ah, and for `ensure_packages` we should pass `provider => 'apt'`.

ppp0:
* true, let's purge the directory before

ppp0:
@njam please review

modules/phantomjs/Puppetfile
```diff
@@ -0,0 +1,10 @@
+mod 'puppetlabs/stdlib',
+  :git => 'git://github.com/puppetlabs/puppetlabs-stdlib.git'
+
+mod 'cargomedia/helper',
+  :git => 'git@github.com:cargomedia/puppet-packages.git',
+  :path => 'modules/helper
```
njam: missing quote

njam:
small syntax error in Puppetfile, rest looks good

ppp0:
trcky one! @njam please review

njam:
Spec:
```
Notice: /Stage[main]/Phantomjs/Helper::Script[install phantomjs]/Exec[exec install phantomjs]/returns: bzip2: (stdin) is not a bzip2 file.
```

ppp0:
retest this please

PR #973, opened 2016-05-02
NVidia Driver

kris-lab:
See https://github.com/cargomedia/puppet-cargomedia/blob/master/docu/bulldog/setup.md#nvidia-drivers

Which one?
```
sudo apt-get install nvidia-346
sudo apt-get install nvidia-352
```

And then?
```
#!/bin/sh
# NOTA BENE: this is expected to run AFTER the reboot from apt-get install nvidia-drivers or
# after removing nouveau modules manually

echo "Step 1: Disable modesetting driver, it breaks Xorg -configure"
mv /usr/lib/xorg/modules/drivers/modesetting_drv.so /usr/lib/xorg/modules/drivers/modesetting_drv.so~

echo "Step 2: Auto-configure with X's own means"
X -configure

echo "Step 3: Move the file back to normal place"
mv -i ~/xorg.conf.new /etc/X11/xorg.conf

echo "Step 4: Restore modesetting driver so it doesn't break the updates"
mv /usr/lib/xorg/modules/drivers/modesetting_drv.so~ /usr/lib/xorg/modules/drivers/modesetting_drv.so

echo "Step 5: Add a custom rule to load NVIDIA's glx xorg module first"
patch -d /etc/X11 <<EOF
--- xorg.conf   2015-10-31 12:46:11.119656923 +0100
+++ xorg.conf-ref       2015-10-31 12:45:26.743234855 +0100
@@ -6,6 +6,8 @@
 EndSection

 Section "Files"
+       ModulePath   "/usr/lib/nvidia-346-updates/xorg/"
+       ModulePath   "/usr/lib/nvidia-352/xorg/"
        ModulePath   "/usr/lib/xorg/modules"
        FontPath     "/usr/share/fonts/X11/misc"
        FontPath     "/usr/share/fonts/X11/cyrillic"
EOF
```


kris-lab:
`/usr/lib/nvidia-346-updates/xorg` does not exists on any of production instances!

kris-lab:
Installation of `nvidia-346-updates` leads to uninstall of `nvidia-346` and opposite!

ppp0:
OMG.. Why I'm reminded of Linus T. ?

kris-lab:
Bulldog testing:
- removed all X11 config files
- run `nvidia-xconfig`
- all works FINE!

Not needed:
- `ModulePath   "/usr/lib/nvidia-346-updates/xorg/"`
- `ModulePath   "/usr/lib/nvidia-346/xorg/"`

njam:
nice :)

kris-lab:
Verified also:
- `apt-get autoremove nvidia-346 nvidia-352`
- added `deb http://archive.ubuntu.com/ubuntu vivid restricted` 
- reinstalled
- run `x_configure` script manually https://github.com/kris-lab/puppet-packages/blob/issue-973/modules/nvidia/templates/configure_x.sh
- test: NOT WORKING
- reboot
- test: WORKING

@njam @ppp0 will test deeper but it seems to be required to reboot after nvidia installation. Will try to unload all from kernel and load manually what needed but not sure what I can do more?

njam:
I think 1 reboot will be a good idea after the initial provisioning anyway. If the full provisioning works with puppet that shouldn't be an issue, we just need to remember it.

kris-lab:
ok, then ready for initial review

kris-lab:
@njam @ppp0 please review

ppp0:
lgtm

njam:
- apt/templates/sources-vivid: Let's add the "restricted" source directly into the first two lines (which have the same codename and mirror, so it can be combined)
- manifests/init.pp: Currently it's mixing arrows and other ordering statements. I would suggest to remove the arrows, and add a `require` or `before` to the `Package` or the `modprobe_blacklist` (will be more flexible).
- configure_x.sh: Remove comment, because it doesn't apply?
- configure_x.sh: Did you check if moving away `modesetting_drv.so` is also needed for `nvidia-xconfig`?

kris-lab:
> configure_x.sh: Did you check if moving away modesetting_drv.so is also needed for nvidia-xconfig?

Indeed looks like not needed anymore. `nvidia-xconfig` ignores any other video modules by leaving `Section "Files"` empty.

@njam please re-review.

njam:
Manual page nvidia-xconfig(1):
> nvidia-xconfig will copy it to /etc/X11/xorg.conf.backup before writing the new configuration

So I think we can remove `configure_x.sh`, and use a plain `exec`? :)

kris-lab:
In general it makes sense and I am gonna to minimize this logic. However the `nvidia-xconfig` takes params '-c' with existing `xorg.conf` file and then makes own changes with `-o` as destination. Both params are default to `/etc/X11/xorg.conf`.

The problem might occur if files already exists with breaking configuration.

kris-lab:
@njam please review

njam:
lgtm

PR #1299, opened 2016-05-03
Add phantomjs to cm in dev

zazabe:


zazabe:
@ppp0 please review

I tried to test it with `bundle exec rake spec:cm debug=true`, but it seems there's no specs to test this module in dev (I can't see the step when `phantomjs` is installed), did I miss something?

njam:
I think there is no spec where we pass `$development => true` to `cm::application`.

zazabe:
@ppp0 could you merge it also if it's okay (I don't have write access to this repo)?

ppp0:
So you don't want to write a spec? 😄

PR #1302, opened 2016-05-04
Update phantomjs installation method

ppp0:
We experienced quite often that our tests fail, due to some unknown temporary glitch at Bitbucket

Could be solved by #885 - doing it locally for now


ppp0:
@njam please review
@zazabe fyi

zazabe:
maybe inline all commands with `&&` + `curl --retry 5` could be enough ?

zazabe:
ok, forget my comment...
as discussed, `bash -e` == `&&` and `curl` could download something else than an achieve, so your implementation sgtm ;)

njam:
So `curl --retry` is not what we want, because it only retries *transient* HTTP errors (5** code).
Let's implement a simple and reusable mechanism to retry any command, so we can reuse it.

Something like:
```sh
retry --tries 3 curl --fail 'http://example.com/foo' | unp
```
Where `--tries` would be optional and default to 3:
```sh
retry curl --fail 'http://example.com/foo' | unp
```

Wdyt?
Either we can implement such a `retry` script in `helper`, or maybe there is an existing program like that we can use?

ppp0:
I would actually retry on transient errors, too - what speaks against it?
I found the one-liner quite elegant and readeable and personally, as so many times, I would send this PR into production and open a new issue about making the retry mechanism reusable

This is my own opinion which I found quite some times to collide. If it is not so urgent, then let's do everything in this PR but I can't guarantee that it will work this week

zazabe:
fyi https://github.com/Medium/phantomjs/issues/506

ppp0:
Thanks @zazabe this saved many people's day

@njam please review

njam:
nice, lgtm

PR #1303, opened 2016-05-04
[security] - Add some policy restrictions to Imagemagick

ppp0:
News go around of a easily exploitable security issue in Imagemagick [CVE-2016–3714](http://www.infoworld.com/article/3065473/security/admins-dont-wait-for-imagemagick-patch.html)


ppp0:
@njam please review
@kris-lab fyi

njam:
Is this policy applied for the php extension? How can it be confirmed?

ppp0:
As I gather from
* https://github.com/mkoppanen/imagick#openmp, 3rd point
* http://php.net/manual/en/imagick.getresourcelimit.php#108776 Sentence:

   >And remember, you can always set the defaults in the policy.xml file on your own server
* http://stackoverflow.com/questions/23111714/memory-limit-reached-creating-a-thumbnail-of-large-image-with-imagick?answertab=votes#tab-top

The `policy.xml` file is being read when invoking ImageMagick, also when this happens through the php extension.

ppp0:
OK, I had to introduce an argument (`$policy_config`) to allow for separate spec testing.
The `policy.xml` file does restrict access. In the spec test's case, to `https` and `mvg`

@njam please review

njam:
- Spec: I guess this should be `echo` instead of `cat`?

ppp0:
Afraid, but a `cat` is needed here, see http://www.tldp.org/LDP/abs/html/here-docs.html#EX71

Did the spec fail?
Can this now be merged?

njam:
You were defining a multiline string containing a PHP script in puppet. Then that PHP script was passed as an argument to `cat` and piped to `php`. So the spec was running this shell script:
```
cat <?php

  $mvg = "push graphic-context
        viewbox 0 0 640 480
        fill 'url(https://127.0.0.1/foo.jpg"| echo 'You are 0wned...'")'
        pop graphic-context";
  $image = new Imagick();
  $image->readImageBlob($mvg);
 | php
```
The result is:
```
line 1: ?php: No such file or directory
You are 0wned...)'
        pop graphic-context
line 3: =: command not found
line 7: syntax error near unexpected token `('
line 7: `  $image = new Imagick();'
```

As you can see the command fails, but the `echo …` was run, so the output contained "0wned". So the spec succeeded with a false positive.

The second spec just checked for the exit code not to be 0, so it succeeded as well, because the command failed.

Both specs did not check anything relevant.

I pushed a fixed version now. Please review.

njam:
The policy works btw, thanks for looking into this!

Please review, then I can deploy it over the weekend.

njam:
Where are the policy contents from? I updated the file with a suggestion from here:
https://bugzilla.redhat.com/show_bug.cgi?id=1332492#c3

I applied the policy in a CM VM and ran the unit tests for `CM_Image`, which succeeded. So I think it should be fine.

Will deploy it now.

ppp0:
uh oh, all looks very good! thanks

PR #1306, opened 2016-05-05
Fix missing augeas double quotas

kris-lab:


kris-lab:
@njam the problem is not in `puppet/ruby` which send proper values to the `augeas`. We have to set explicitly that we wanna have this quotas for non string values.

njam:
Could we instead pass `value => '"true"'`?

kris-lab:
@njam yes:) please re-review

njam:
lgtm

PR #1311, opened 2016-05-06
Re-add basic imagick spec

njam:
@ppp0 please review

I didn't notice this spec was removed in #1303, and I removed some assertions from the other spec.


ppp0:
lgtm!

PR #1277, opened 2016-05-09
Satis: send cron output to root, not to satis user

ppp0:
For example there's a user "satis" and it has a cron job. If that cron job fails an email will be sent to the user, and forwarded to "satis@cargomedia.ch".

To receive such emails we should forward it to "root".

Maybe something like http://www.postfix.org/postconf.5.html#luser_relay

cc @kris-lab @ppp0

### DIFF
--- a/modules/satis/manifests/repo.pp
+++ b/modules/satis/manifests/repo.pp
@@ -1,4 +1,6 @@
-define satis::repo ($content) {
+define satis::repo (
+  $content,
+  $cron_environment = 'MAILTO=root') {
 
   require 'satis'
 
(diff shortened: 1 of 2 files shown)


ppp0:
As discussed, we want to redirect mails originally meant for *existing* users, not *non-existing*

The easiest solution is to add a `MAILTO` env var to the cron job which contains the final recipient on a per-case basis

@njam please review

njam:
@kris-lab could you review?

kris-lab:
Looks good to me. 

I am just wondering... shouldn't it be a default behaviour for each cron job?

ppp0:
This would make sense. Afaics, here (https://github.com/cargomedia/puppet-packages/blob/master/modules/satis/manifests/repo.pp#L27) is the only cron job so far that is not running for root. So it's the quickest workaround. 

We probably could define smth like
```pp
Cron {
   environment => 'MAILTO=root',
}
```

PR #1314, opened 2016-05-11
Add basic docker module

njam:


njam:
Paired with @tomaszdurka

PR #1315, opened 2016-05-11
Require fluentd for janus

njam:


njam:
Paired with @tomaszdurka

PR #1316, opened 2016-05-11
Introduce "docker" case for daemon

njam:


njam:
Paired with @tomaszdurka

PR #1317, opened 2016-05-11
Notify daemon instead of service

njam:
Depends on https://github.com/cargomedia/puppet-packages/pull/1316


njam:
Paired with @tomaszdurka

PR #1322, opened 2016-05-18
Implement janus-cluster member - register end-point

tomaszdurka:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1274

Not fully standard to https://github.com/cargomedia/janus-cluster

### DIFF
--- a/modules/janus_cluster/manifests/member.pp
+++ b/modules/janus_cluster/manifests/member.pp
@@ -1,0 +1,25 @@
+define janus_cluster::member (
+  $cluster_manager_url,
+  $webSocketAddress,
+  $data = { },
+) {
+
+
+  $register_data = {
+    id               => $name,
+    webSocketAddress => $webSocketAddress,
+    data             => $data,
+  }
+
+  file { '/usr/bin/register-janus-cluster-member':
+    content => template('janus_cluster/register_member.sh.erb'),
+    owner   => 'root',
+    group   => 'root',
+    mode    => '755',
+  }
+  ~>
+
+  daemon { 'janus-cluster-member':
+    binary => '/usr/bin/register-janus-cluster-member',
+  }
+}
--- a/modules/janus_cluster/spec/member/manifest.pp
+++ b/modules/janus_cluster/spec/member/manifest.pp
@@ -1,0 +1,13 @@
+node default {
+
+  janus_cluster::member { 'node-id':
+    cluster_manager_url => 'http://cluster-manager',
+    webSocketAddress    => 'ws://node-address',
+    data                => {
+      rtpbroadcast => {
+        role                => 'repeater',
+        upstream            => 'upstream-id',
+      }
+    }
+  }
+}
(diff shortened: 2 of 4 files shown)


tomaszdurka:
@kris-lab wanna review?

tomaszdurka:
retest this please

tomaszdurka:
@cargomedia/devops can anyone tell me why I am getting this error about invalid variable for `JSON.generate`? This runs correct on my local environment. 

Is puppet on CI somehow different than our local ones and possibly passes variables to templates in different way?

Thanks.

njam:
Should not, those commands run with `bundle exec` and the `Gemfile.lock` is committed!
I can reproduce locally:
```sh
$ bundle exec rake syntax
---> syntax:manifests
---> syntax:templates
rake aborted!
JSON::GeneratorError: only generation of JSON objects or arrays allowed
/Users/reto/Projects/puppet-packages/.bundle/vendor/ruby/2.1.0/gems/json_pure-1.8.3/lib/json/common.rb:223:in `generate'
/Users/reto/Projects/puppet-packages/.bundle/vendor/ruby/2.1.0/gems/json_pure-1.8.3/lib/json/common.rb:223:in `generate'
modules/janus_cluster/templates/register_member.sh.erb:6:in `<main>'
/Users/reto/Projects/puppet-packages/.bundle/vendor/ruby/2.1.0/gems/puppet-syntax-1.4.1/lib/puppet-syntax/templates.rb:17:in `block in check'
/Users/reto/Projects/puppet-packages/.bundle/vendor/ruby/2.1.0/gems/puppet-syntax-1.4.1/lib/puppet-syntax/templates.rb:13:in `check'
/Users/reto/Projects/puppet-packages/.bundle/vendor/ruby/2.1.0/gems/puppet-syntax-1.4.1/lib/puppet-syntax/tasks/puppet-syntax.rb:51:in `block (2 levels) in initialize'
Tasks: TOP => syntax => syntax:templates
(See full trace by running task with --trace)
```

Are you sure you're using the correct bundler environment? Maybe run `bundle install`?

kris-lab:
- mode: maybe `755` => `0755 `?
- unify naming: `webSocketAddress` => `websocket_address `?

As we discussed I am not sure about name `janus_cluster::member` - this logic sounds more like helper registration or similar. Maybe "member_helper" makes more sense? However I am not sure if we extend this resource in the future. @tomaszdurka wdyt?

tomaszdurka:
Maybe we should reconsider naming and modules in general.

You probably want to name this helper, as you don't see this small module as "janus_cluster::member", but just addition. That's probably my bad as of naming `janus-cluster` this way. `janus-cluster` should be used rather as concept not single piece of software. Originally nodejs app was called `janus-cluster-manager` which probably is better. Than currently implemented module would be called `janus-cluster-manager::member/subscriber` which makes more sense. 

Possibly we could add a wrapper class for what we try to add withing `janus::role::cluster...` and instead have it within `janus_cluster::member` or even `janus_streaming_cluster::member`.

kris-lab:
closing in favor of https://github.com/cargomedia/puppet-packages/pull/1204

PR #1327, opened 2016-05-23
Cron job verbosity for satis

ppp0:


ppp0:
To find clues about what makes satis cron job fail, let's grab the output and attach it to cron's output

@njam agree?

njam:
Is there a simple way to store it in a bash variable?

ppp0:
A multi-line output you mean? - a single tmpfile not good?

njam:
I thought about something like `foo=$(command) || echo "${foo}"`. So we don't need a file.
But if it's not easily possible then file is good as well!

PR #1332, opened 2016-05-24
[Satis] Clean-up cron output in case of errors

ppp0:


ppp0:
@kris-lab please review

ppp0:
pair reviewd w @kris-lab

kris-lab:
As discussed, let try this approach!

PR #1337, opened 2016-05-26
Install pulseaudio service with optional modules

kris-lab:


kris-lab:
@ppp0 not sure about that... wdyt? See usage https://github.com/cargomedia/puppet-cargomedia/pull/1214/files#diff-a9afb52964c3a1204622d9848dd524bcR27

ppp0:
lgtm
what you don't like?

kris-lab:
I had a thought that maybe it should be proxied from `pulseaudio::service` to `pulsaudio` which would install it. However it is overcomplicated, so if you agree I can merge this.

ppp0:
I agree it'a bit of a a mess. We configure modules in the service. Otoh, service is non-standard handling (normal is autorespawn and all automagic).. 

May be the naming is unfortunate and should reflect our application somehow? I don't know. Let's see how far we get with this now

PR #1293, opened 2016-05-30
cm::reverse_proxy: add spec with SSL for the proxy

ppp0:
To prevent bugs like the one fixed here: https://github.com/cargomedia/puppet-packages/pull/1292

Remove non-https support?


ppp0:
OK - this was a big one, lots of debugging

New:
* Frontends, ie reverse-proxies are SSL-only
* The original idea of having SSL-only towards upstream was scrapped, ie it can be configured, default is http (the status-quo). I think it would add a lot of overhead where we actually don't need it, e.g. on private networks or when we use reverse-proxy as SSL endpoint

@njam fyi

@kris-lab please review

kris-lab:
@ppp0 could you have a look on failing specs? I will review shortly

ppp0:
Let's wait for CI. @kris-lab please review

modules/cm/manifests/reverse_proxy.pp
```diff
@@ -10,13 +11,14 @@
 
   $upstream_options_defaults = {
     name        => 'reverse-proxy-backend',
-    members     => ['localhost:443'],
-    ssl         => true,
+    members     => ['localhost:80'],
     header_host => '$host',
+    ssl         => false,
```
kris-lab: I thought we wanna enforce SSL to be used? What is the point of keeping this `ssl` key in the hash and later on switch the protocol based on `$ssl`?

modules/cm/manifests/reverse_proxy.pp
```diff
@@ -10,13 +11,14 @@
...
   }
 
   $upstream_opts = merge($upstream_options_defaults, $upstream_options)
   $upstream_name_real = $upstream_options[name] ? { default => $upstream_opts[name], undef => $upstream_options_defaults[name] }
+  $ssl = $upstream_options[ssl] ? { default => $upstream_opts[ssl], undef => $upstream_options_defaults[ssl] }
```
kris-lab: I think it is not needed?

modules/cm/manifests/reverse_proxy.pp
```diff
@@ -27,36 +29,35 @@
...
-    }
+  nginx::resource::vhost{ "${name}-https-redirect":
+    listen_port         => 80,
+    server_name         => $hostnames,
+    location_cfg_append => [
+      'return 301 https://$host$request_uri;',
+    ],
+  }
+
+  $proto = $ssl? {
```
kris-lab: Is it needed as well?

modules/cm/manifests/reverse_proxy.pp
ppp0: This is the *upstream* configuration. I think it is not always desirable to have ssl on the backlink. One prominent example being the use case of reverse proxy as a SSL endpoint

modules/cm/manifests/reverse_proxy.pp
ppp0: Well, I was confused by line 20.. Why was it needed there? Is `merge` not reliable or doesn't it know what to do when a default shall be overwritten?

modules/cm/manifests/reverse_proxy.pp
kris-lab: Sorry, I was confused indeed. It makes sense then!

modules/cm/manifests/reverse_proxy.pp
kris-lab: Please ignore my comment. I read this wrong.

kris-lab:
I would suggest we pair review this, wdyt?

modules/cm/manifests/reverse_proxy.pp
ppp0: Let's read from ` $upstream_opts[ssl]`

modules/cm/manifests/reverse_proxy.pp
ppp0: As discussed, let's keep the default from before (`ssl => true`) otherwise we end up having to reconfigure production in many places

ppp0:
@kris-lab one final look please

modules/cm/manifests/reverse_proxy.pp
kris-lab: I would remove `$ssl` - it is not needed.

modules/cm/manifests/reverse_proxy.pp
```diff
@@ -27,36 +29,32 @@
...
+  nginx::resource::vhost{ "${name}-https-redirect":
+    listen_port         => 80,
+    server_name         => $hostnames,
+    location_cfg_append => [
+      'return 301 https://$host$request_uri;',
+    ],
   }
 
-  $proto = $upstream_opts[ssl] ? { true => 'https', false => 'http' }
+  $proto = $ssl? { false => 'http', default => 'https'}
```
kris-lab: What about we remove `$ssl` and keep like it was before?

ppp0:
@kris-lab please re-review

modules/cm/manifests/reverse_proxy.pp
```diff
@@ -10,7 +11,7 @@
 
   $upstream_options_defaults = {
     name        => 'reverse-proxy-backend',
-    members     => ['localhost:443'],
+    members     => ['localhost:80'],
```
kris-lab: @ppp0 is it ok we have `ssl=true` as upstream and `localhost:80` as defaults?

kris-lab:
I think it looks good. I have added small note.

modules/cm/manifests/reverse_proxy.pp
ppp0: AARGH, of course NOT, thank you

ppp0:
Build did not fail, Jenkins did on tear down
```sh
Finished in 30 mins 58 secs.
Success! 15 specs run, 0 failures (83 examples, 0 failures)

vagrant halt --force
==> Debian-7: Forcing shutdown of VM...
FATAL: org.acegisecurity.providers.UsernamePasswordAuthenticationToken cannot be cast to org.jenkinsci.plugins.GithubAuthenticationToken
java.lang.ClassCastException: org.acegisecurity.providers.UsernamePasswordAuthenticationToken cannot be cast to org.jenkinsci.plugins.GithubAuthenticationToken
[..]
Setting status of 569e175d997aa4e82dd706db37855b5854cc1eb5 to FAILURE with url http://ci.cargomedia.ch:8080/job/cargomedia-puppet-packages-pr/196/ and message: 'Build finished. '
Finished: FAILURE
```

PR #1340, opened 2016-05-30
Remove imagetragick spec

njam:
This has been fixed in debian packages.

@ppp0 please review


ppp0:
How has this been fixed? By changing the defaults? Otherwise I think it still makes sense to limit the powers of a libmagick client if they're not needed

njam:
Yes it's fixed in debian packages. The spec fails now therefore. I'm keeping the policy though.

njam:
Ah, how: https://security-tracker.debian.org/tracker/CVE-2016-3714

ppp0:
LGTM

PR #1342, opened 2016-05-31
[Fluentd] - Gem dependency requires ruby 2.0

ppp0:
We're having errors on puppet runs, not sure on which nodes, wheezy only?

Looks like a `fluentd` dependency wants ruby 2.0

````

*** [XXX] Error: Could not update: Execution of '/usr/bin/gem install --no-rdoc --no-ri fluentd' returned 1: ERROR:  Error installing fluentd:
*** [XXX] strptime requires Ruby version ~> 2.0.
*** [XXX] Error: /Stage[main]/Fluentd/Ruby::Gem[fluentd]/Package[fluentd]/ensure: change from ["0.12.26", "0.12.25", "0.12.24", "0.12.23", "0.12.22"] to 0.14.0 failed: Could not update: Execution of '/usr/bin/gem install --no-rdoc --no-ri fluentd' returned 1: ERROR:  Error installing fluentd:
*** [XXX] strptime requires Ruby version ~> 2.0.
```

### DIFF
--- a/modules/fluentd/manifests/init.pp
+++ b/modules/fluentd/manifests/init.pp
@@ -1,7 +1,14 @@
 class fluentd {
 
+  # Version 0.14.0 breaks wheezy
+  # Dep strptime requires Ruby version ~> 2.0
+  $fluentd_version = $::facts['lsbdistcodename'] ? {
+    'wheezy' => '0.12.26',
+    default  => latest,
+  }
+
   ruby::gem { 'fluentd':
-    ensure => latest,
+    ensure => $fluentd_version,
   }
 
   file { '/etc/fluentd':


ppp0:
@njam I don't like to do this, really, but since we're going to move to Jessie soonish, I'd say this is a viable workaround? Wdyt?

@kris-lab please review

kris-lab:
For records:

> What version of Ruby does fluentd support?
Fluentd v0.10 or v0.12 works on 1.9.3 or later. Since v0.14, 2.1 or later.
http://docs.fluentd.org/articles/faq

ppp0:
Hm, very interesting :) - so it's official...

kris-lab:
Ok, I agree in this case seems to be only way to go.

kris-lab:
@ppp0 as the provider is `gem` can we maybe allow to install any version of  `~0.12.0`

ppp0:
Hm, doesn't seem to work

```sh
gem install fluentd --version '~0.12'
ERROR:  While executing gem ... (Gem::Requirement::BadRequirementError)
    Illformed requirement ["~0.12"]

gem install fluentd --version '~0.12.0'
ERROR:  While executing gem ... (Gem::Requirement::BadRequirementError)
    Illformed requirement ["~0.12.0"]
```

😢

ppp0:
I would merge this. On Jessie, `fluentd` has been automatically upgraded to 0.14 without fuzz. On Wheezy, we have all nodes failing. So nothing can happen really

ppp0:
@kris-lab is it ok?

njam:
sgtm

kris-lab:
lgtm

ppp0:
specs failing because this is the fix

PR #1347, opened 2016-06-01
fluent-plugins breaking on Wheezy

ppp0:
Continuation of https://github.com/cargomedia/puppet-packages/pull/1342

next gem affected

```sh
Error: Could not update: Execution of '/usr/bin/gem install --no-rdoc --no-ri fluent-plugin-sentry' returned 1: Building native extensions.  This could take a while...
ERROR:  Error installing fluent-plugin-sentry:
	strptime requires Ruby version ~> 2.0.
Error: /Stage[main]/Fluentd::Plugin::Sentry/Ruby::Gem[fluent-plugin-sentry]/Package[fluent-plugin-sentry]/ensure: change from absent to latest failed: Could not update: Execution of '/usr/bin/gem install --no-rdoc --no-ri fluent-plugin-sentry' returned 1: Building native extensions.  This could take a while...
ERROR:  Error installing fluent-plugin-sentry:
	strptime requires Ruby version ~> 2.0.
````

### CODE REVIEW

**Comment by ppp0** [comment]
@kris-lab please review

**Comment by ppp0** [comment]
It looks to me as the plugins try to install the *latest* version of gem fluentd - which breaks on wheezy.
Regardless of `fluentd` being installed (and compatible according to the plugin's gemspecs). Opened a PR

**Comment by njam** [comment]
As discussed - shouldn't a gemspec requirement *without* a version use the version already installed on the system? Is the older fluentd gem already installed when the plugin gets installed?

**Comment by ppp0** [comment]
The tests I did where on a machine with fluentd already installed in an older flavour. Also see this very PR: it requires module `fluentd` - What amounts to the gem `fluentd` being installed first *in the correct version* (and it fails...)

**Comment by ppp0** [comment]
There's even a gem for that: https://github.com/joonty/pessimize :)

**Comment by ppp0** [comment]
So it seems that installing a plugin leads to an error, but it gets installed anyway.

**Comment by ppp0** [comment]
Investigtation w/ @kris-lab showed that *on a second run* all is good. So `gem install` fails the fist time, but it actually completes the job (of installing a plugin)

**Comment by ppp0** [comment]
As discussed, let's `require` fluentd class and remove the `require =>` arg

**Comment by ppp0** [comment]
Problem: spec failing for Wheezy. It would need a 2nd puppet run *with box kept* (which is not possible atm)

Possible solutions
* `exec {}` gem install with `|| true` so that it doesn't fail
* ignore both manifest and spec for Wheezy (needs a switch based on distro)

**Comment by ppp0** [comment]
Still failing with plugin fluent-plugin-sentry v0.0.5 (which contains https://github.com/y-ken/fluent-plugin-sentry/pull/7)..

There is an option to `gem` (`--conservative`, see http://guides.rubygems.org/command-reference/#gem-install) - But unfortunately, it seems our version (on Wheezy, quite old) is buggy in this respect (https://github.com/rubygems/rubygems/issues/929)

**Comment by kris-lab** [comment]
As discussed the spec fails so let's work it around for the moment. We are gonna migrate to the Debian-8 and the problem will vanish itself.

As researched the problem appears only for the first run of puppet.

**Comment by ppp0** [comment]
@kris-lab Please review

**Comment by kris-lab** [comment]
retest this please

**Comment by kris-lab** [comment]
@ppp0 could you check specs?

**Comment by ppp0** [comment]
fluentd service failing with
```
2016-06-10 11:57:52 +0000 [error]: config error file="/etc/fluentd/fluent.conf" error="Unknown filter plugin 'record_modifier'. Run 'gem search -rd fluent-plugin' to find plugins"
```

will check

ppp0:
obsoleted / not the solution

PR #1355, opened 2016-06-02
Add "after" and "before" parameter to daemon

ppp0:


ppp0:
@kris-lab please review

Ideas on how to 100% make sure that the dependency is honoured by `systemd` and not by puppet?

kris-lab:
- modules/daemon/manifests/init.pp: has the `$after` to be default?

I don'y fully understand what we test in the spec:) could you explain a bit?

ppp0:
As discussed with @njam, let's introduce `before`, too..

The tests:
* foo service executes a one-shot program which in turn creates a file in `/tmp`
* my-program "daemon" expects the file to be there to run in a loop

kris-lab:
- `before` sounds good!
- test: ok, I agree

njam:
For the records - as discussed yesterday: Systemd's ordering instructions (Before/After) apply to when starting multiple units at the same time. So for example at boot. I think it will be difficult to test that in a acceptance spec, I also think it might not be necessary.

ppp0:
@kris-lab @njam anyone up for a review?

kris-lab:
@ppp0 quick question: did we decide to ignore before/after configuration on the `sysvinit`? Maybe we should make it more clear?

ppp0:
We had a look at https://wiki.debian.org/LSBInitScripts on how to do this in init scripts. Unfortunately, there is some sort of `after` (defining which service has to be started before, see `X-Start-Before` and `should-start`) but not `before`.

While it would be nice to have a puppet define that abstracts from the init system, we discussed that it's too much of an effort given that we're going to phase out Wheezy in the near future

I can make a comment in the code

kris-lab:
As discussed let's close this thread. It does not solve the problem.

closing in favor of https://github.com/cargomedia/puppet-packages/issues/1359

PR #1359, opened 2016-06-15
Make pulseaudio service capable to restart system services after reload/restart

kris-lab:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1152

Roughly workflow:
image


kris-lab:
@ppp0 have a look

We have at least 2 options:
- just proxy `post_command` from the service down to daemon
- allow to pass list of services to the `pulseaudio::service` then evaluate the list for existence, status and then restart

I am up to simplify this process and allow to pass any value.

cc @njam

njam:
How about doing it automatically `if defined('lightdm')` - I guess it's always desired behaviour.

kris-lab:
Well, I added this changes cuz I partially agree. It is always desired by us :) because it is needed only with `KIOSK` mode. Regular `X` session e.g. `gnome` will auto publish `Xsession` into `pulseaudio` on start!

kris-lab:
@ppp0 wdyt?

ppp0:
There are alternatives to `lightdm`, see http://askubuntu.com/questions/545373/where-can-i-find-alternative-greeters-for-lightdm

On to other hand, this is a specialized module for our purposes. But then, does it fit into this (public) repo or should it be configurable (with a parametrized desktop manager/greeter to depend on).. 

It seems to function and fulfill what we want it to do, so it kinda lgtm

njam:
Isn't it the window manager we're interested in?
Seems there's no simple way to restart or even determine the active window manager: http://askubuntu.com/questions/72549/how-to-determine-which-window-manager-is-running

Anyway "lightdm" should be sufficient for this module I guess.

ppp0:
sure, the window/display manager, this is a better link https://wiki.mageia.org/en/Display_Managers#Other_Display_Managers

ppp0:
A window manager is again something different: https://wiki.archlinux.org/index.php/Window_manager#List_of_window_managers

kris-lab:
My question is: should we make configurable the "lightdm restart" behaviour? or we allow that any mode "gnome", "kiosk" and so on will be restarted if `pulseaudio` restarts? This is our assumption that the display manager is not able to handle this case (like the kiosk mode is not able to do it by itself).

ppp0:
What happens in non-kiosk mode? Does it hurt if pulseaudio restarts lightdm?

kris-lab:
It hurts a lot imo. Restart of `lightdm` will restart `DM` like e.g. `gnome`. Then it will restart the `Xsession` and all applications which runs there, agree?

njam:
But it might still be needed? Because if you restart pulseaudio then gnome will *not* know about the new process and probably not be able to play audio?

kris-lab:
Yep, let's keep restart behaviour.

kris-lab:
@ppp0 @njam final review... or you OK to merge?

njam:
lgtm

modules/pulseaudio/spec/service_with_lightdm/spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+describe 'pulseaudio::service' do
```
ppp0: May be change the spec test's name here?

ppp0: `describe 'pulseaudio::service::service-with-lightdm' `

ppp0:
I took some notes, rest LGTM

ppp0:
wait for CI then it's mergeable imo

kris-lab:
Will merge during production downtime...

ppp0:
👍

PR #1358, opened 2016-06-15
Make sure daemon's monit entry for systemd does not output non-ASCII characters

ppp0:
We should disable output completely or disable non-ASCII characters.

### DIFF
--- a/modules/daemon/templates/monit.systemd.erb
+++ b/modules/daemon/templates/monit.systemd.erb
@@ -1,2 +1,2 @@
-check program <%= @name %> path "/bin/systemctl status <%= @name %>"
+check program <%= @name %> path "/bin/systemctl status <%= @name %> >/dev/null"
 	if status != 0 then alert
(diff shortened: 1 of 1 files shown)


ppp0:
The only way to make this sure is to redirect the output to `/dev/null`. May be we could redirect also stderr's output? But if `systemctl status` fails, it's interesting if monit can provide us with details

@kris-lab agree?

kris-lab:
Did you try to redirect all to `/dev/null` and see what is the exit code? I think `monit` checks the exit code and not the stdout/err output?

ppp0:
I did, and yes, monit checks the exit code - which is unaffected by stdout/stderr output

My question is: shall we dump all output to /dev/null or just stdout?

kris-lab:
It depends if `stderr` contains non-ASCII characters. If yes, then both should to the null/

ppp0:
I don't know how to get `systemctl` to output something to stderr...
I guess we can safely dump both stderr and stdout.

We're really interested in the exit code only here

@kris-lab please review

kris-lab:
ok, lgtm, works for me.

PR #1362, opened 2016-06-15
Debug monit systemctl status

ppp0:


ppp0:
self-merging due to triviality

PR #1366, opened 2016-06-16
Remove spl_types php extension as a cm dependeny

njam:
@fvovan please review

Reference: https://github.com/cargomedia/cm/pull/2202


fvovan:
lgtm

PR #1365, opened 2016-06-16
Swallow non-ASCII characters from systemctl status

ppp0:
Continuation of https://github.com/cargomedia/puppet-packages/pull/1358

It looks like monit does not like output redirections on scripts it's calling. 

Idea: write a wrapper script

```sh
#!/bin/bash

exit $(systemctl status foo 2>&1 >/dev/null)
```

something like that...


ppp0:
@kris-lab please review

kris-lab:
So the problem is with STD redirections yes?

What about we introduce `systems-service-status` custom script which takes argument of service name? Then we would not have so many custom files there. Is it possible to pass an argument with monit check?

ppp0:
Agree, agree!

@kris-lab please review

kris-lab:
- rename `${title}-status` with `service-status`?

Should we also prefix this `service-status` with some cargomedia name?

ppp0:
I'm on solving this now.. `daemon` being a define, we can't use `service-status` as file name (will lead to duplicate declaration)

ppp0:
Bad news, it looks like monit does not accept arguments passing to external scripts 👎 
````
root@debian-8-amd64-plain:/home/vagrant# monit summary
/etc/monit/conf.d/my-program:1: Program does not exist: ''/usr/local/bin/service-status my-program''
/etc/monit/conf.d/my-program:1: Program does not exist: ''/usr/local/bin/service-status my-program''
AssertException: File '/usr/local/bin/service-status my-program' does not exist
```

ppp0:
From https://mmonit.com/monit/documentation/monit.html#PROGRAM-STATUS-TESTING

>You can also send parameters with the program:

> check program list-files with path "/bin/ls -lrt /tmp/"
       if status != 0 then alert
>Arguments to the program or script is a sequence of whitespace separated strings. In the above example the strings '-lrt' and '/tmp/' are arguments to the program '/bin/ls'. If arguments are used, it is recommended to use quotes " to enclose the string, otherwise, if no arguments are used, quotes are not needed.

>Notes: If the program is a script, the interpreter is required in the first line. The program or script must also be executable.


TL;DR: trying with double quotes

kris-lab:
Hmm `Monit::Entry[$title]` is virtual resource in that case. Maybe we have to introduce ordering like e.g. https://github.com/cargomedia/puppet-packages/blob/master/modules/apt/manifests/update.pp#L30

kris-lab:
In general I agree. As I mentioned during discussion I am not a huge fun `systemctl_status` as a part of `monit`.

What about moving `systemctl_status` into `systemd` module? I think it will be cleaner because at the end of the day the `systemctl_status` has nothing to do with `monit`?

@njam wdyt?

njam:
Agree, monit doesn't need to know about this. I would also say this should be part of the `daemon` module.
Other thoughts:
- Let's use a proper file extension (`.erb`) for templates. Current ones (`.debian`, `.systemd`) will confuse any editor.
- /usr/local/bin/systemctl-status: Would be nice to *not* prefix this command with `systemctl`, to keep the bash autocompletion for `systemctl` useful.

ppp0:
This was part of `daemon` before doing some refactoring.. (see https://github.com/cargomedia/puppet-packages/pull/1365#issuecomment-226472682). I have the feeling we're spending lots of effort in something we're going to ditch anyway in a couple of months (mis-using monit for plain monitoring systemd processes)

njam:
Agree that using monit for this is not perfect. But I don't see a viable alternative right now (including providing the ability to turn off alerting during maintenance).
If there was one, then we could even replace monit *today* for jessie.

kris-lab:
@ppp0 is it still relevant?

ppp0:
I'd say fortunately not!

PR #1371, opened 2016-06-22
Mongodb for Jessie

ppp0:


ppp0:
@kris-lab please review - let's wait for CI

kris-lab:
retest this please

ppp0:
depends on https://github.com/cargomedia/puppet-packages/pull/1376

kris-lab:
Still failing. @ppp0 let's try to resolve before asap!

ppp0:
depends on #1381

ppp0:
test will fail without #1381

ppp0:
@njam please review (let's hope CI passes). This PR is required by #1254

njam:
- Remove monit from specs (as it's not relevant to this module anymore)?
- mongodb/manifests/core/mongod.pp: I would change those `before => Daemon` to `require => …` on the deamon. Because the files don't care about the daemon, but rather the daemon cares about the files. This way the daemon has a easily readable list of resource it needs to function.
 - Same for mongodb/manifests/core/mongos.pp
- mongodb/manifests/core/mongos.pp: Why can't we use the `start_on_create` of *daemon* instead of starting manually here?

Rest lgtm

ppp0:
>mongodb/manifests/core/mongos.pp: Why can't we use the start_on_create of daemon instead of starting manually here?

This has cost @kris-lab and me already a lot of headache. We found out that somehow services are started upon definition (hence the introduction of `start_on_create`). By notifying the `Exec` *after* definition of the service, we ensure that services are not abruptly restarted just after creation (e.g. due to config change notifications)

njam:
With "services are started upon definition" I guess you mean that the starting of the service is a resource declared *inside* the `Daemon` resource?
The config files notify the service like this:
```
File['/some/config'] ~> Service['mongodb']
```

While System::Unit has:
```
File['/unit/description'] ~> Exec['start mongodb']
Service['mongodb']
```

Wouldn't the correct solution be to add an ordering of `Service -> Exec['start']` to Systemd (and sysvinit)?
This way we will *only* start the service *after* it has been defined (if needed).
Compare:
```
sudo puppet apply -e 'notify{"foo":} ~> exec{"/etc/init.d/ntp start": unless=>"/etc/init.d/ntp status"} notify{"foo2":} ~> service{"ntp": provider=>"debian"}' --debug
```
With:
```sh
sudo puppet apply -e 'notify{"foo":} ~> exec{"/etc/init.d/ntp start": unless=>"/etc/init.d/ntp status", require=>Service["ntp"]} notify{"foo2":} ~> service{"ntp": provider=>"debian"}' --debug
```

Let me know if you want to pair?

njam:
To summarize we would need:
```
Service['mongodb'] -> Exec['start mongodb']
```

kris-lab:
As discussed with @ppp0 the problem is not with `mongo`, but rather with `daemon` implementation which seems to restart the service after create. For mongo this is killing behaviour in scenario, when only single router is active. It creates distributed lock (then dies - restarted) which is never removed by any active router.

njam:
I think it's because of the order of resources as discussed with Philipp. Did you try the proposed change?
```
Service['mongodb'] -> Exec['start mongodb']
```

ppp0:
I just did

```
   Error: Failed to apply catalog: Found 10 dependency cycles:
      (Exec[systemctl start mongod_config1] => Systemd::Unit[mongod_config1] => Service[mongod_config1] => Exec[systemctl start mongod_config1])
      (Exec[systemctl start mongod_config2] => Systemd::Unit[mongod_config2] => Service[mongod_config2] => Exec[systemctl start mongod_config2])
      (Exec[systemctl start mongod_config3] => Systemd::Unit[mongod_config3] => Service[mongod_config3] => Exec[systemctl start mongod_config3])
      (Exec[systemctl start mongod_rep1_arbiter] => Systemd::Unit[mongod_rep1_arbiter] => Service[mongod_rep1_arbiter] => Exec[systemctl start mongod_rep1_arbiter])
      (Exec[systemctl start mongod_rep1_db1] => Systemd::Unit[mongod_rep1_db1] => Service[mongod_rep1_db1] => Exec[systemctl start mongod_rep1_db1])
      (Exec[systemctl start mongod_rep1_db2] => Systemd::Unit[mongod_rep1_db2] => Service[mongod_rep1_db2] => Exec[systemctl start mongod_rep1_db2])
      (Exec[systemctl start mongod_rep2_arbiter] => Systemd::Unit[mongod_rep2_arbiter] => Service[mongod_rep2_arbiter] => Exec[systemctl start mongod_rep2_arbiter])
      (Exec[systemctl start mongod_rep2_db1] => Systemd::Unit[mongod_rep2_db1] => Service[mongod_rep2_db1] => Exec[systemctl start mongod_rep2_db1])
      (Exec[systemctl start mongod_rep2_db2] => Systemd::Unit[mongod_rep2_db2] => Service[mongod_rep2_db2] => Exec[systemctl start mongod_rep2_db2])
      (Exec[systemctl start mongos_router1] => Systemd::Unit[mongos_router1] => Service[mongos_router1] => Exec[systemctl start mongos_router1])
    Try the '--graph' option and opening the resulting '.dot' file in OmniGraffle or GraphViz
```

ppp0:
unit should probably not depend on service.. hmm trying

ppp0:
pushing some code to run specs on ci overnight

modules/sysvinit/manifests/script.pp
```diff
@@ -1,34 +1,26 @@
...
-  }
 
   Service <| title == $name |> {
     enable    => true,
     provider  => 'debian',
     subscribe => File["/etc/init.d/${name}"],
   }
+  ->
+
+  exec { "/etc/init.d/${name} start":
```
njam: Won't this execute on every puppet run? So it will start the service whenever it's not running (e.g. during app deployment)?

ppp0:
Retrying with `refreshonly => true` and `Service ~> Exec`

ppp0:
By disabling (aka commenting out) the step, all specs will succeed. I would open a follow-up ticket to look into the issue and move forward with our Jessie migration

ppp0:
Grrr 3 spec failures:
1x network issue
2x spec tests for sysvinit

Hope now it will succeed!

ppp0:
@kris-lab @njam yipppieee - please review

modules/mongodb/spec/cluster/3-app2.pp
```diff
@@ -19,13 +19,13 @@
...
-  ->
-
-  mongodb_collection { 'bar':
-    ensure    => present,
-    database  => 'dummydb',
-    shard     => true,
-    shard_key => '_id',
-    router    => 'localhost:27017'
-  }
+#  ->
```
kris-lab: should it be re-added?

ppp0: If we do, this particular spec won't pass (after a whooping 15hrs or so)... 
Interstingly, spec `cluster:with-auth` succeeds

kris-lab: Ahh I thought that with changes to `daemon` we do not restart service more than once? It should work if service is started and kept alive.

modules/mongodb/manifests/core/mongos.pp
```diff
@@ -13,6 +13,14 @@
   $daemon = 'mongos'
   $instance_name = "${daemon}_${name}"
 
+  if $::facts['service_provider'] == 'debian' {
```
njam: remove this?

njam:
- unit.pp / script.pp: I would still notify the "start" *exec* from the *file*, but ensure it runs after the *service* by adding ordering to `Service`. Because now we notify from a `Service` which might not be realized at all (it is done outside in *daemon*).

ppp0:
retest this please

ppp0:
@kris-lab @njam please review again

njam:
- core/mongod.pp: Probably "cluster-key-file" should also be created before daemon (like in mongos)?
- cluster/3-app2.pp: What about this spec? This causes the hours long build? Can you confirm that the services are not restarted after the first start?

ppp0:
* makes sense!
* I would say let's forget this spec and merge for the moment. Let's fix this at a later stage.. The services are not restarted anymore after creation, at least there is no more mentioning of that in the logs.

The `mongos_router` is caught in a weird loop:

```
2016-07-31T19:24:50.724+0000 [Balancer] distributed lock 'balancer/debian-8-amd64-plain:27017:1469992872:1804289383' acquired, ts : 579e50822e7b22c5edac9a0b
2016-07-31T19:24:50.734+0000 [Balancer]  ns: dummydb.bar going to move { _id: "dummydb.bar-_id_MinKey", ns: "dummydb.bar", min: { _id: MinKey }, max: { _id: 0 }, version: Timestamp 1000|0, versionEpoch: ObjectId('579e4fcf2e7b22c5edac99ec'), lastmod: Timestamp 1000|0, lastmodEpoch: ObjectId('579e4fcf2e7b22c5edac99ec'), shard: "rep1" } from: rep1 to: rep2 tag []
2016-07-31T19:24:50.734+0000 [Balancer] moving chunk ns: dummydb.bar moving ( ns: dummydb.bar, shard: rep1:rep1/localhost:27001,localhost:27002, lastmod: 1|0||000000000000000000000000, min: { _id: MinKey }, max: { _id: 0 }) rep1:rep1/localhost:27001,localhost:27002 -> rep2:rep2/localhost:27006,localhost:27007
2016-07-31T19:24:50.737+0000 [Balancer] moveChunk result: { ok: 0.0, errmsg: "migration already in progress" }
2016-07-31T19:24:50.737+0000 [Balancer] balancer move failed: { ok: 0.0, errmsg: "migration already in progress" } from: rep1 to: rep2 chunk:  min: { _id: MinKey } max: { _id: 0 }
2016-07-31T19:24:50.800+0000 [Balancer] distributed lock 'balancer/debian-8-amd64-plain:27017:1469992872:1804289383' unlocked. 
```

ppp0:
As discussed: Cluster has one part commented out. Cluster-with-auth works. Will debug at a later stage as cluster on a single node is not a real use case

njam:
agree

PR #1374, opened 2016-06-27
Janus: network tuning

kris-lab:
As investigated with Chris we should:

Increase the network receive buffer with sysctl:
- net.core.rmem_max = 32000000
- net.core.rmem_default = 32000000

This will help about "RcvbufErrors" and subsequently dropped frames. We can monitor that with netstat: https://github.com/cargomedia/bipbip/issues/154

### DIFF
--- a/modules/janus/manifests/common.pp
+++ b/modules/janus/manifests/common.pp
@@ -11,6 +11,13 @@ class janus::common (
     system => true,
   }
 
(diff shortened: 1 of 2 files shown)


kris-lab:
@ppp0 please review

ppp0:
look good

PR #1381, opened 2016-06-30
Optionally prevent autostart of daemons at creation time

ppp0:


ppp0:
@kris-lab kindly review, please

ppp0:
retest this please

ppp0:
@njam please review

njam:
- Instead of introducing a new variable `$unless_start_on_create` could we just use `! $start_on_create`?
```sh
$ sh -c '(! false) && echo jo'
jo
```

ppp0:
It's true, but otoh
* the variable introduced is local, the resulting code is still far from unreadeable?
* It's a way of *sanitise* the input.. One could pass other values than a boolean and have them executed.

I know it's an abstract use case but it's going against some security basics

njam:
I found it rather a bit confusing with an inverse variable of the input, and the comment explaining the logic. Having a *logical not* operator would be more understandable to me.
But it's fine like this too.
lgtm

PR #1383, opened 2016-07-04
Add new prefix /resources for cm cdn-origin vhost

njam:
@fauvel please review

The idea is to introduce a new prefix `/resources`, which would be the *only* prefix for all js/css/image resources mid-term.
We can already use it for the new response you introduced.


fauvel:
Make sense! Lgtm.

PR #1384, opened 2016-07-07
Make coturn compatible with Debian Jessie

ppp0:


ppp0:
This was almost too easy, @njam wanna review? :)

njam:
lgtm

I would vote to drop Ubuntu support when not needed, to reduce maintenance and CI duration.

PR #1388, opened 2016-07-20
Chromium bleeding-edge

ppp0:
For https://github.com/cargomedia/janus-gateway-js/pull/67 it is necessary to have a newer version of chromium than the one that comes with the deb package

### DIFF
--- a/modules/chromium/manifests/init.pp
+++ b/modules/chromium/manifests/init.pp
@@ -1,9 +1,29 @@
-class chromium {
+class chromium($build = undef) {
 
   require 'apt'
 
-  package { 'chromium-browser':
-    ensure   => present,
-    provider => 'apt',
+  if $build != undef {
+
+    ensure_packages([
+      'libpangocairo-1.0',
+      'libnss3',
+      'libcups2',
+      'libgconf2-4',
+      'libatk1.0-0',
+      'libasound2',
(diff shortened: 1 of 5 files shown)


ppp0:
@kris-lab please review

ppp0:
@kris-lab please review

modules/chromium/manifests/init.pp
```diff
@@ -1,9 +1,29 @@
...
+      'libatk1.0-0',
+      'libasound2',
+      'libgtk2.0-0',
+      'libxss1',
+      'libxtst6'
+    ], { provider => 'apt' })
+
+    helper::script { 'install chrome browser':
+        content => template("${module_name}/chromium-installer/install.sh.erb"),
+        unless  => 'ls /usr/bin/chrome',
```
kris-lab: @ppp0 should it be possible to install any version of chrome or if any is installed then we stop installation?

kris-lab: Maybe use something like
```
$ chromium-browser --version
Chromium 48.0.2564.82 Ubuntu 15.04
```

kris-lab:
As discussed let's open follow up ticket and try to improve the installation a bit: check for the version `chromium-browser --version` and download the installer with more trusted place:)

PR #1396, opened 2016-07-20
Correct wrong unless clause in Chromium bleeding-edge

ppp0:
Introduced in #1388


ppp0:
@kris-lab please review

Right now, we need to manually uninstall a present chromium (or just remove the `/usr/bin/chromium-browser` file) if we want to switch version. Updates when installed via `deb` package is working as before

kris-lab:
lgtm

PR #1397, opened 2016-07-20
Check permissions on chromium-browser tree

ppp0:
@vogdb reports this https://github.com/cargomedia/janus-gateway-js/pull/67#issuecomment-233971551

### DIFF
--- a/modules/chromium/templates/chromium-installer/install.sh.erb
+++ b/modules/chromium/templates/chromium-installer/install.sh.erb
@@ -4,7 +4,8 @@ curl -sL "https://www.googleapis.com/download/storage/v1/b/chromium-browser-snap
 rm -rf /usr/local/chrome-linux
 unzip -d /usr/local chrome-linux.zip
 chmod -R +r /usr/local/chrome-linux
-chmod +x  /usr/local/chrome-linux /usr/local/chrome-linux/chrome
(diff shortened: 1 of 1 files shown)


ppp0:
@kris-lab please review

modules/chromium/templates/chromium-installer/install.sh.erb
```diff
@@ -4,7 +4,8 @@ curl -sL "https://www.googleapis.com/download/storage/v1/b/chromium-browser-snap
 rm -rf /usr/local/chrome-linux
 unzip -d /usr/local chrome-linux.zip
 chmod -R +r /usr/local/chrome-linux
-chmod +x  /usr/local/chrome-linux /usr/local/chrome-linux/chrome
+chmod +x /usr/local/chrome-linux
+find /usr/local/chrome-linux -type f -perm -700 -exec chmod +x {} \;
```
kris-lab: so the idea is to make binary files executable or all files with `0700`?

ppp0: The idea is to make all files that are marked *owner-executable* world executable, so that every user can execute

kris-lab: ok then

kris-lab:
lgtm

PR #1399, opened 2016-07-22
Add needrestart watcher for services

kris-lab:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1299


kris-lab:
@ppp0 please review

ppp0:
lgtm

kris-lab:
As discussed we have some options agains current implementation (which is the most flexible and extensible):
- DEB: provide simple script we assumes service name (e.g. `janus_*`) and restarta all services it matches
- DEB: provide more complicated script which runs `needrestart` internally. It would allow to restart only janus instances which are actually affected by upgrade (e.g. only `audio` or only `streaming` plugin). It will require to provide name of service to match (e.g. `janus_*`)

ppp0:
I would vote to integrate this into the plugin's deb. On the other hand, the restart script itself might need `needrestart` - we do customize the installation of janus ourselves - by adding path and service name prefixes - so that a simple postinstall script might have a hard time finding all services that depend upon the plugin

njam:
Interesting!
`needrestart/Puppetfile` should require apt.

njam:
I thought about adding this to the DEB package first too. Then the post-install script would need to detect all services that are affected by the package's files. Might be difficult to accomplish.

Maybe it's better to leave the DEB as is, and add this "needrestart" for every cluster service define of *janus* and *cm-janus*?
Is the `needrestart` execution fast enough, so running it dozens of times after every `dpkg` execution doesn't have a negative impact?

kris-lab:
Agree to keep it out of DEBs. Running `needrestart` for many services should not be an issue, but single run takes 2-3sec. I believe the hooks from `apt/dpkg`will run all in parallel.

If you agree on proposed solution in PR, I will add more specs and benchmark resource use for multi/parallel `needrestart` run.

njam:
> `DPkg::Post-Invoke`, which is invoked after every execution of dpkg (by an apt tool, not manually);
http://unix.stackexchange.com/questions/226993/whats-the-difference-between-dpkgpost-invoke-and-dpkgpost-invoke-success

This means that many operations, like `apt-get update` or `apt-get autoclean` etc will be taking 2-3 seconds longer, correct? This might negatively affect our [apt related tasks in pulsar-conf-cargomedia](https://github.com/cargomedia/pulsar-conf-cargomedia/blob/master/recipes/linux/apt.rb). Could you test it?

ppp0:
retest this please

kris-lab:
Please ignore 2-3 sec processing time. I have double checked on production and in testing VM. Seems to be quick now. Maybe it was because of my VMs (not sure atm).

Check/restart timestamps from the spec:
```
-rw-r--r--  1 root    root       0 Jul 27 13:38 my-program1-check-stamp-1469626692.791008602
-rw-r--r--  1 alice   alice      0 Jul 27 13:38 my-program1-start-stamp-1469626692.803372311
-rw-r--r--  1 root    root       0 Jul 27 13:38 my-program2-check-stamp-1469626692.872864890
-rw-r--r--  1 alice   alice      0 Jul 27 13:38 my-program2-start-stamp-1469626692.880663127
-rw-r--r--  1 root    root       0 Jul 27 13:38 my-program3-check-stamp-1469626692.944699733
-rw-r--r--  1 alice   alice      0 Jul 27 13:38 my-program3-start-stamp-1469626692.951248412
```

It means there is practically no delay.

kris-lab:
@njam it seems to work reliable. I didn't check with `pulsar`, mainly because it runs very quick. There might be a scenario that it takes some time if we run `needrestart` for 20-30 times, but I don't see how could it affect the other commands/apps in the pipeline.

please review

kris-lab:
retest this please

njam:
- needrestart/manifests/service.pp: Currently `$title` is the name of the *process* to look for and the name of the *service* that should be restarted. It wasn't so obvious for me. I would suggest to introduce a parameter `$service_name` instead (could still default to `$title`)?
- needrestart/spec/service/spec.rb: The spec is checking that the daemon hasn't created any files (which it does only every 1 second). Why? What does that prove?
 - Best would probably be if the daemon process would load a library file, then we would delete that file, then run "apt-get upgrade", then verify that the process was restarted? Not sure if it's feasible though.
- needrestart/templates/needrestart: Let's make this grep more strict, so it doesn't match substrings?
- needrestart/templates/needrestart: Let's put the service name into quotes
- needrestart/templates/needrestart: Maybe we should check first if a service by that name exists. And if it doesn't we can show an appropriate message "Process 'foo' needs restart bug can't find service by that name"?
 - Btw what happens for the "apt-get upgrade" execution if such a post command fails?

kris-lab:
> The spec is checking that the daemon hasn't created any files (which it does only every 1 second). Why? What does that prove?

File is created only once, after start. The loop does 1 sec delay forever.

Agree, it is not the right way to go. The idea was that the service creates file after restart, so if `dpkg:post-invoke` run `needrestart` checks and it matches the service name, then restart should happen and file should be created (see that files are removed before test run). However this test has too many edge cases to be reliable in current version.

kris-lab:
> Btw what happens for the "apt-get upgrade" execution if such a post command fails?

I have tried with corrupted/failing script. It throws error and continue normal operation...
```bash
(...)
Processing triggers for initramfs-tools (0.120+deb8u2) ...
update-initramfs: Generating /boot/initrd.img-3.16.0-4-amd64
I am here
/usr/local/bin/needrestart-my-program1: line 5: makeerrorforme: command not found
Scanning processes...                                                                                                                                                                
Scanning candidates...                                                                                                                                                               
Scanning kernel images...                                                                                                                                                            
Restarting services using systemd...
```

kris-lab:
> Best would probably be if the daemon process would load a library file, then we would delete that file, then run "apt-get upgrade", then verify that the process was restarted? Not sure if it's feasible though.

The only solution working for me is:
- setup the service with "binary" as ruby application
- modify content of "binary" application what is detectable for `needrestart` (when I used bash script instead of ruby script the changes of file has been not detected by `needrestart`)

kris-lab:
@njam please re-review

njam:
- templates/needrestart: Let's add file extension ".sh.erb", it helps the IDE?
- templates/needrestart: How about passing the service name as an argument to this script, then only 1 version of the script is needed, and it's easier to re-use it.
- templates/needrestart: I'm sorry because I suggested this change. But maybe we don't need to check if the unit exists? systemctl already does it and we will see the error in apt output?
```
$ systemctl restart foo
Failed to restart foo.service: Unit foo.service failed to load: No such file or directory.

```

njam:
lvgtm

kris-lab:
I assume your last comment is not by accident and code is ready to merge.

PR #1400, opened 2016-07-26
Service start depends on service definition (Implies: Revert "Optionally prevent autostart of daemons at creation time")

ppp0:
Reverts cargomedia/puppet-packages#1381

As discussed in https://github.com/cargomedia/puppet-packages/pull/1371#issuecomment-231007992 we better attain this goal by changing the dependency/notification structure a tiny bit


ppp0:
@njam please review - this basically reverts the changes introduced in #1381 and ensures that a service gets started as discussed in https://github.com/cargomedia/puppet-packages/pull/1371#issuecomment-231007992

njam:
We should not do `ensure    => running`. It will start the service on every puppet run, even if it should be stopped (e.g. during a release).

But just adding the suggested dependency should do the trick I think?

ppp0:
solved in #1404

PR #1404, opened 2016-07-27
Remove optional start of daemon, rearrange dependencies

ppp0:


ppp0:
@kris-lab please review

kris-lab:
Can you have a look on specs?

ppp0:
retest this please

ppp0:
This is interesting.
**spec succeded, jenkins fails to post back**

```
Finished in 27 mins 33 secs.
Success! 25 specs run, 0 failures (140 examples, 0 failures)

vagrant halt --force
==> Debian-8: Forcing shutdown of VM...
==> Ubuntu-15.04: Forcing shutdown of VM...
==> Debian-7: Forcing shutdown of VM...
Setting status of 309d997bba0b52e330a836c6204e27c6ca18f865 to SUCCESS with url http://ci.cargomedia.ch:8080/job/cargomedia-puppet-packages-pr/128/ and message: 'Build finished. '
Could not update commit status of the Pull Request on GitHub.
java.io.IOException: <html>
<head><title>405 Not Allowed</title></head>
<body bgcolor="white">
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

@kris-lab ok to merge?

modules/sysvinit/spec/coredump/1-manifest.pp
```diff
@@ -13,6 +13,5 @@
 
   sysvinit::script{ 'foo':
     content   => template('sysvinit/spec/script/foo'),
-    before    => Service['foo'],
```
kris-lab: So we remove `before => Service['foo']` but we are not 100% sure why?

kris-lab:
I am not 100% sure why it helps. I am aware of dependencies cycle, however why it happens now? In general if all services are defined properly (read: by `daemon{}`) then all should be fine imo.

modules/systemd/manifests/unit.pp
```diff
@@ -1,36 +1,29 @@
...
-    refreshonly => true,
+    notify => Exec['systemctl daemon-reload'],
   }
 
   Service <| title == $name |> {
     enable    => true,
     provider  => 'systemd',
     subscribe => File["/etc/systemd/system/${name}.service"],
   }
+  ~>
```
ppp0: @kris-lab this is the new dependency we establish that produces dep cycles whenever you start to require service elsewhere in the specs

ppp0: I took `unit.pp` (systemd) but the same applies to `script.pp` (sysvinit)

ppp0:
But I get your point.. It might be that we have this dependency defined somewhere in other modules.. which would wreck our setup!

kris-lab:
Yes, this is what I am not sure 100%. Let's run all specs then?

ppp0:
@kris-lab on second thought, this should only affect modules that invoke `sysvinit::script` directly. Which is neither recommended, nor useful. All service/daemons should be defined via `daemon` otherwise you lose the abstraction sysvinit/systemd.
We had to write spec tests for `sysvinit`, obviously, but it's not a use case otherwise

ppp0:
I found two places where we use `sysvinit:script`:
* https://github.com/cargomedia/puppet-packages/blob/master/modules/mms/manifests/agent/monitoring.pp#L35
* https://github.com/cargomedia/puppet-packages/blob/master/modules/mms/manifests/agent/backup.pp#L32

which should be rewritten anyway for jessie.. but still, they are expected to fail

ppp0:
Copied @njam from https://github.com/cargomedia/puppet-packages/pull/1371#issuecomment-235914592:
>unit.pp / script.pp: I would still notify the "start" exec from the file, but ensure it runs after the service by adding ordering to Service. Because now we notify from a Service which might not be realized at all (it is done outside in daemon).

ppp0:
@njam please review

ppp0:
Oops, I need to fix https://github.com/cargomedia/puppet-packages/pull/1404#issuecomment-235912569 too

njam:
- unit.pp / script.pp: Instead of a separate ordering, how about just adding `after => Service` to the `exec`?
- script.pp: formatting

rest lgtm

ppp0:
retest this please

ppp0:
retest this please

ppp0:
retest this please

PR #1405, opened 2016-07-28
Integrate janus::server with needrestart::service

kris-lab:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1299
Depends on https://github.com/cargomedia/debian-packages/issues/132

### DIFF
--- a/modules/janus/manifests/server.pp
+++ b/modules/janus/manifests/server.pp
@@ -110,7 +110,9 @@ define janus::server (
     ],
   }
 
-  needrestart::service { $instance_name:
+  if $::facts['lsbdistcodename'] == 'jessie' {
+    needrestart::service { $instance_name:
+    }
   }
 
   if $instance_name != 'janus' {
--- a/modules/janus/manifests/server.pp
+++ b/modules/janus/manifests/server.pp
@@ -110,6 +110,9 @@ define janus::server (
     ],
   }
 
+  needrestart::service { $instance_name:
+  }
+
   if $instance_name != 'janus' {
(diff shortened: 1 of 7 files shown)


kris-lab:
retest this please

njam:
How about we do it only for jessie? I understand for now it's needed for janus and cm-janus, which will be fully on jessie soon.

kris-lab:
How would you do it then? We need still compatibility with Wheezy... or you mean to apply `needrestart::service{}` only for `jessie` and ignore for `wheezy`?

njam:
Hm, yes why not an `if`. Once we're done with wheezy we can then grep for these IFs and delete them…

kris-lab:
@njam please review

Recently the `janus-gateway-rtpbroadcast@0.0.22` has been released. It is not applied to production so I will test current logic with the manual upgrade.

modules/janus/manifests/server.pp
```diff
@@ -110,6 +110,11 @@
     ],
   }
 
+  if $::facts['lsbdistcodename'] == 'jessie' {
```
njam: I would suggest to do the opposize (`!= 'wheezy'`?), will be easier to search and remove…

kris-lab: OK, however we will search by e.g. `lsbdistcodename` to be on the safe side.

njam:
true, lgtm

Did you test all of it together?

kris-lab:
I did some extended tests in my local VM/jessie. Seems to work perfectly.

Let's see how it goes on the production.

PR #1414, opened 2016-08-01
Debian-8: "cron" compatibility

kris-lab:


kris-lab:
@ppp0 please have a look

ppp0:
Would you like some "fun"? -> https://blog.higgsboson.tk/2013/06/09/use-systemd-as-a-cron-replacement/

ppp0:
lgtm

kris-lab:
Agree to replace, but really should we do it now:) ?

ppp0:
nonono, it was kind of a joke

modules/cron/manifests/init.pp
```diff
@@ -8,12 +8,7 @@
...
   ->
 
-  service { 'cron':
-    enable => true,
-  }
-
-  @monit::entry { 'cron':
-    content => template("${module_name}/monit"),
-    require => Service['cron'],
+  daemon {'cron':
```
kris-lab: @ppp0 do you think we should use `daemon` instead of `service`? Maybe we should leave the `service` as abstraction (as it is now) and only add the condition `if system == Wheezy` then install `monit`?

ppp0: Not sure! Did you cross-check with `/etc/init.d/cron` ?  I would opt for migrating this to `daemon`. Did you notice some pitfalls?

kris-lab: I didn't notice any troubles, but I could miss something. Let's merge and see what happens.

kris-lab:
Problems:
- if `cron` package is installed it starts automatically
- then fork which creates `/var/run/crond.pid` file

If we use custom `daemon` with name `cron` (required, to not have duplicated service files from DEB) then we expect `/var/run/cron.pid` to be in place (expected by monit). Problems:
- `cron` starts and fork by ignoring `start-stop-daemon` `--pidfile` configuration

Solutions:
- we can pass `-f` to `daemon` which prevents forking on `cron`, but it is not restarting `cron` installed/started by `apt` under `sysvinit` (it works for `systemd)`
- any other solutions ???

wdyt @ppp0

kris-lab:
Antoher solution would be to add `if OS=Wheezy` then use old approach (service + monit) and any other would use `daemon` resource?

kris-lab:
@ppp0 approach like this e.g.
```
 package { 'mongodb-org':
    provider => 'apt',
    require  => [ Apt::Source['mongodb'] ]
  }
  ->
  service { 'mongod':
    ensure => stopped,
    enable => false,
  }
  ->
```
will not work, because the assumption with approach above is that the new service would have a different name `mongo_XXXX`, `janus_XXX` which is not a case for `cron`.

ppp0:
Then may be this?

```pp
 package { 'cron':
    ensure   => present,
    provider => 'apt',
  }
  ~>

  exec { '/etc/init.d/cron stop':
    command   => '/etc/init.d/cron stop',
    path      => ['/bin','/usr/bin'],
    refreshonly => true,
  }
```

ppp0:
@kris-lab I pushed my changes. Basically:
* stop daemon at install time
* define daemon with puppet, remove unneeded pidfile `/var/run/crond.pid` (created by `/usr/bin/cron`)
* add spec to check that pidfile isn't there

let's see if it passes CI

ppp0:
Hm - otoh, package `cron` is very likely to be already there. so it won't notify the Exec to stop it

modules/cron/manifests/init.pp
```diff
@@ -6,14 +6,17 @@
...
+    refreshonly => true,
   }
 
-  @monit::entry { 'cron':
-    content => template("${module_name}/monit"),
-    require => Service['cron'],
+  daemon { 'cron':
+    binary       => '/usr/sbin/cron',
+    args         => '-f',
+    post_command => 'rm -f /var/run/crond.pid',
```
kris-lab: If we use `-f` then the file `/var/run/crond.pid` will be created only once after `apt-get install`, yes?

kris-lab:
Hmm will not notify... however it works now, wondering why?

kris-lab:
Well, I think we can skip problem of Jessie/Ubuntu as from the beginning it worked on both. The problem was only with Wheezy yes?

ppp0:
I'm still investigating why the spec 
```` 
describe file('/var/run/crond.pid') {
      it { should_not be_file }
   }
```
succeeds but the file is there! doesn't make sense to me

ppp0:
@kris-lab yes it works, but to make sure the service gets stopped before we install our daemon-script, I had to resort change things a bit. I'm not sure if this is a correct change (having `Package -> Exec['stop cron']`)

ppp0:
@kris-lab please review

kris-lab:
@ppp0 as discussed, after long discussion and many many hack we tried... I think only dirty workarounds have left. Will do some...

kris-lab:
So I added not so bad workaround. I simply kill all running `cron` only if new service is created. It makes the initial state clean imo.

@ppp0 wdyt?

ppp0:
I think I like but CI does not completely

modules/cron/spec/default/spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    it { should be_enabled }
+    it { should be_running }
+  end
+
+  describe command('ls /var/run/crond.pid') do
+    its(:exit_status) { should_not eq 0 }
+  end
+
+  describe command('sudo ps aux | pgrep cron | wc -l') do
+    its(:content) { should eq 1 }
```
ppp0: its(:stdout) would work

modules/cron/spec/default/spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    it { should be_installed }
+  end
+
+  describe service('cron') do
+    it { should be_enabled }
+    it { should be_running }
+  end
+
+  describe command('ls /var/run/crond.pid') do
+    its(:exit_status) { should_not eq 0 }
```
ppp0: away with this 😄

kris-lab: It works locally. However I have replaced `:content` with `:stdout`

kris-lab:
@ppp0 please review

modules/cron/spec/default/spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    it { should be_installed }
+  end
+
+  describe service('cron') do
+    it { should be_enabled }
+    it { should be_running }
+  end
+
+  describe command('sudo ps aux | pgrep cron | wc -l') do
+    its(:stdout) { should match /1/ }
```
ppp0: may be `/^1$/ ` ?

ppp0:
hmhm, lgtm.. see comment . we will survive 😄

modules/cron/spec/default/spec.rb
kris-lab: agree would be better. I was expecting value: 0|1|2 tbh

njam: There's also http://serverspec.org/resource_types.html#process

kris-lab: Makes sense!

modules/cron/templates/monit
```diff
@@ -1,3 +1,3 @@
 check process cron with pidfile /var/run/crond.pid
-	start program = "/etc/init.d/cron start"
-	stop  program = "/etc/init.d/cron stop"
+  start program = "/etc/init.d/cron start"
```
njam: This file can probably be removed

kris-lab: true!

ppp0: oh, my bad! of course!

kris-lab: done https://github.com/cargomedia/puppet-packages/pull/1431

PR #1415, opened 2016-08-01
Debian-8: "elasticsearch" compatibility

kris-lab:


kris-lab:
@ppp0 could you have a look on "environment_file" logic for `daemon`? I am not sure how to solve this best for `elasticsearch`. thx

kris-lab:
retest this please

kris-lab:
@ppp0 I was about to introduce next `daemon` param called `environment_file`. In general it makes sense to support that, however for `elasticsearch` I would also propose this solution https://github.com/cargomedia/puppet-packages/pull/1415/commits/0a929b76791ac7d6ac4737ff8d9e115f9b33bb13 - wdyt?

kris-lab:
@ppp0 please review. notes:
- after DEB install, the new service is stopped (affects systemd and sysvinit)
- default `heap_size` is set to `1g` - as documented by elasticsearch
- removed environment file and moved configuration to the daemon (mainly because of systemd and future wise plans)

kris-lab:
@njam could you please review?

njam:
- init.pp: Set the default for the heap size in the class parameter list?
- init.pp: RESTART_ON_UPGRADE: I guess this is [read by a script in the package](https://github.com/elastic/elasticsearch/blob/2539d94bc95f651971b1c23db395caa603ce4b0f/distribution/src/main/packaging/scripts/postinst#L74). So we should still write a default file, with this in it.
- install.sh: Could we use the apt repo instead? -> https://www.elastic.co/guide/en/elasticsearch/reference/master/deb.html

kris-lab:
@njam thx for all tips. please review

njam:
- elasticsearch/manifests/common.pp: Usually we have a "::common" class if the main entry point is a *define* (then the class represents the "common" resources of all defines). Here it's not needed and could just be inlined. Or if we want to have it in a separate file then maybe call it `elasticsearch::package`?

Rest lgtm

kris-lab:
Hmm I did it in purpose. Unfortunately I get this:
```
(Exec[Add deb signature key for elasticsearch] => Exec[apt_update] => Package[openjdk-7-jre-headless] => Class[Java::Jre_headless] => Class[Elasticsearch] => Apt::Source[elasticsearch] => File[/etc/apt/sources.list.d/elasticsearch.list] => Exec[apt_update])
```
not sure what exactly rise this dependency cycle.

kris-lab:
I prefer to keep it as cleaner as `elasticsearch::package`.

@njam please re-review

njam:
lgtm

PR #1254, opened 2016-07-27
Make "cm" module compatible with Debian Jessie

ppp0:
part of https://github.com/cargomedia/puppet-cargomedia/issues/1157

PR #1402, opened 2016-08-01
cron job output in php5/jessie

ppp0:
part of #1254

On our Jessie nodes with cm we get sysadmin mail every 30min:

````
Cron <XXX> [ -x /usr/lib/php5/sessionclean ] && /usr/lib/php5/sessionclean
Inbox
x 
Cannot load Zend OPcache - extension already loaded
```

### DIFF
--- a/modules/php5/manifests/extension/opcache.pp
+++ b/modules/php5/manifests/extension/opcache.pp
@@ -22,10 +22,21 @@ class php5::extension::opcache (
       require => Class['php5'],
       before  => Php5::Config_extension['opcache'],
     }
-  }
 
-  php5::config_extension { 'opcache':
-    content => template("${module_name}/extension/opcache/conf.ini"),
+    php5::config_extension { 'opcache':
+      content => template("${module_name}/extension/opcache/conf.ini"),
+    }
+
+  } else {
+
+    file { "/etc/php5/mods-available/opcache.ini":
+      ensure  => file,
+      content => template("${module_name}/extension/opcache/conf.ini"),
+      owner   => '0',
+      group   => '0',
+      mode    => '0644',
+    }
+
   }
 
   Php5::Fpm::With_opcache <||>
--- a/modules/php5/manifests/extension/opcache.pp
+++ b/modules/php5/manifests/extension/opcache.pp
@@ -29,7 +29,7 @@ class php5::extension::opcache (
 
   } else {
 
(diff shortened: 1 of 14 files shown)


ppp0:
@njam please review. We have two possibilities here:
* Delete `/etc/php5/{cli,other_sapis}/conf.d/05-opcache.ini` before everything in `php5::extention::opcache` (but one needs to go over all possible/installed SAPIs, e.g. cli, fpm, cgi, etc.)
* (as of here) - Skip the step `php5enmod` for just this extension

ppp0:
@njam please hold - not completed yet

ppp0:
@njam please review - I revised my opinion, also tried `php5dismod` which did not work :(

modules/php5/manifests/extension/opcache.pp
```diff
@@ -22,6 +22,16 @@
       require => Class['php5'],
       before  => Php5::Config_extension['opcache'],
     }
+  } else {
+
+    exec { 'Remove all symlinks to opcache':
+      command     => 'find /etc/php5 -name 05-opcache.ini -delete',
+      path        => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
+      provider    => shell,
+      unless      => 'find /etc/php5/ -name 05-opcache.ini | grep -q 05-opcache.ini',
```
ppp0: This construct is needed, because the exit code does not state if any files were found or not. Excerpt from manpage:

>EXIT STATUS
       find  exits  with  status  0  if all files are processed successfully, greater than 0 if errors occur.   This is deliberately a very broad
       description, but if the return value is non-zero, you should not rely on the correctness of the results of find.

njam:
Oh.
Let's add a spec assertion to verify it's not loaded twice (see e.g. `php5/spec/extension/apcu/spec.rb`)?

ppp0:
Good idea
@njam please review

ppp0:
or @kris-lab you can review, too

kris-lab:
`phpdismod` seems to very smart from what I read.

lgtm

ppp0:
>phpdismod seems to very smart from what I read.

😄 depends... May be it works for extensions that have been enabled by the user. It did not work for the built-in opcache.ini

kris-lab:
ahah sorry, should be "phpdismod seems to **BE NOT** very smart"

PR #1243, opened 2016-08-02
Port mysql module to jessie

ppp0:
Depends on https://github.com/cargomedia/puppet-packages/pull/1422


ppp0:
I found a unit file and an accessory script for pre/post in https://launchpad.net/ubuntu/+source/mysql-5.7/5.7.13-0ubuntu0.16.04.2 - more specifically in [mysql-5.7_5.7.13-0ubuntu0.16.04.2.debian.tar.xz](https://launchpad.net/ubuntu/+archive/primary/+files/mysql-5.7_5.7.13-0ubuntu0.16.04.2.debian.tar.xz)

ppp0:
depends on #1422

ppp0:
@cargomedia/devops I would really prefer to not touch the existing Wheezy deployment! To accomplish this, I'm resorting to an otherwise ugly `if os==Wheezy..then..else`

I hope you agree. The reason is that mysql's init script is highly...well, scripted, so that it becomes an error-prone task to try to mimic this in our `daemon` define.

The new unit+helper script are copied from a PPA (Ubuntu dev repo) for mysql 5.7 and they seems to work

ppp0:
@kris-lab please review

ppp0:
Either way, moving our infra to mysql5.5 won't be a matter of just applying this. We need to plan the transition. I suggest we order a new set of servers, deploy them from scratch, bring the database to run there, do the switch and later decommission the old db servers. To be discussed in detail

njam:
Regarding migration: I suggest to upgrade a single slave first and verify that it runs fine for a few days. Then upgrade the other slaves and the master.
We can either add a new slave, but we could also replace an existing one like "maintenance db".

ppp0:
@kris-lab please review

ppp0:
Failure was
```
rake aborted!
JSON::GeneratorError: only generation of JSON objects or arrays allowed
```

hm?

ppp0:
retest this please

ppp0:
Quoting @njam 
>Regarding migration: I suggest to upgrade a single slave first and verify that it runs fine for a few days. Then upgrade the other slaves and the master.
We can either add a new slave, but we could also replace an existing one like "maintenance db".

Good idea! I'd add another slave and if things go well, promote it to "maintenance"

kris-lab:
@ppp0 taking over from here

kris-lab:
@njam could you have a look? As we discussed, it requires extra script for proper operations. It can be removed in version `5.7+` (included in package). Also as discussed initially, we do introduce switch (if) for Wheezy/Jessie to stay on the safe site!

njam:
- mysql-systemd-start.sh.erb: Add a note with a reference where it is coming from?

Rest lgtm

kris-lab:
Affected nodes with upgrade
```
 ** [out :: janus4.cargomedia.net (10.71.140.10)] Process 'mysql-proxy'               Not monitored
 ** [out :: janus5.cargomedia.net (10.68.236.116)] Process 'mysql-proxy'               Not monitored
```
currently not monitored

PR #1421, opened 2016-08-03
logrotate issues can't find process to kill

ppp0:
New Sysadmin mails on Jessie servers:
```
Cron <XXX> test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.daily )
```

```
password
to root 
/etc/cron.daily/logrotate:
logrotate_script: 2: kill: No such process

error: error running non-shared postrotate script for /var/log/mongodb/mongos_router.log of '/var/log/mongodb/mongos_router.log '
run-parts: /etc/cron.daily/logrotate exited with return code 1
```

### DIFF
--- a/modules/mongodb/manifests/core/mongos.pp
+++ b/modules/mongodb/manifests/core/mongos.pp
@@ -75,6 +75,6 @@ define mongodb::core::mongos (
   logrotate::entry { $instance_name:
     path              => "/var/log/mongodb/${instance_name}.log",
     rotation_newfile  => 'create',
-    postrotate_script => "kill -USR1 $(cat /var/run/${instance_name}.pid)",
+    postrotate_script => "kill -USR1 $(cat /var/run/${instance_name}.pid) 2>&1 >/dev/null || true",
   }
 }


ppp0:
So may be mongodb crashed when sending SIGUSR1? (see https://jira.mongodb.org/browse/SERVER-11087?focusedCommentId=787813&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-787813)

Either way, we will move away from log files and their rotation eventually (daemon's dumps to STDERR/STDOUT will be captured by journald)

I checked the logs and they're all there, properly rotated. Let's ignore an eventual failure as in http://stackoverflow.com/a/8396266

ppp0:
@kris-lab please review

kris-lab:
@ppp0 as discussed let's keep it open and observe the problem. I would say the problem should be fixable on Mongo module level. Muting this problem is not the way to go.

ppp0:
Investigation shows the pidfile `/var/run/mongos_router.pid` on `janus5` to be stale.. likely due to transition to systemd.

We need other ways to identify the PID in systemd.. `pgrep` or, more robust `systemctl show mongos_router -p MainPID | cut -d '=' -f 2`

Mongo expects SIGUSR1 after log rotation. `systemctl` only sends SIGTERM, SIGINIT and SIGKILL, so we still need `kill`

njam:
Do we even need logrotate with systemd? Why not use the Journal, I understand it's how it's meant to be used and will solve all of this for us?

ppp0:
I completely agree that logrotate is kinda obsoleted. I think we need to get rid of log files at all in order to properly transfer control to journald. Otherwise we end up with a logfile that grows indefinitely and journald that acts as a front-end. In sum: if we write log files, we should rotate them

njam:
How about we disable logging for mongodb and use the journal?
Then we add an exception for wheezy: enable logging and add logrotation. Once wheezy is not needed anymore, this exception can be removed.

ppp0:
@njam closing this, solving in #1430, reopen if you agree

PR #1425, opened 2016-08-03
mysqlproxy for Jessie

ppp0:
It depends on #1243 changes...


ppp0:
@kris-lab please review (#1243 to be merged with master and later master pulled here first)

kris-lab:
ok, let's review and merge stuff one-by-one. The code with all this pulls is kind of unreadable.

ppp0:
I'm very sorry. The only commit which is relevant to this PR is https://github.com/cargomedia/puppet-packages/pull/1425/commits/7fac6636e1b733ec2386dc3d45d94e34079f92a0

ppp0:
The dependency (#1243) is really **only for the spec tests* to succeed

kris-lab:
@ppp0 taking over from here

kris-lab:
@njam wanna have a look?

njam:
- mysql_proxy/manifests/init.pp: I would in general try to avoid using `->` and `~>`, because it's much easier to refactor code that uses `require`/`notify` etc (except maybe for tests), wdyt?
- mysql_proxy/spec/init/spec.rb: Monit check can be removed here, since this is now handled by the "daemon" module, which has tests for it. We check that the port is listening, which should be enough here.

kris-lab:
@njam please re-review

njam:
lgtm

kris-lab:
Deployment will affect `www[3,4,5]`, `search2` and `janus[2,4,5]`. Some cleanup might be required.

PR #1433, opened 2016-08-09
Add modules for phpkafka

alexispeter:
Modules for https://github.com/arnaud-lb/php-rdkafka and required https://github.com/edenhill/librdkafka


alexispeter:
Please review @kris-lab

kris-lab:
- php5/Puppetfile: add `librdkafka`
- modules/librdkafka/manifests/init.pp - WARNING: double quoted string containing no variables on line 7
- `librdkafka`: is there reason we do not use DEB for this library, version constraint? https://packages.debian.org/jessie/librdkafka-dev
- modules/librdkafka/spec/init/spec.rb: use `file()` to check if file exists http://serverspec.org/resource_types.html#file

alexispeter:
I just used 0.9.1 because the php-binding has the same version. I can try if it works with 0.8.5 if this is an issue.

kris-lab:
Using provided DEBs from official repository is just simpler for maintenance and the puppet implementation is also more readable. If not working then lets stick to the manual installation (like is now).

alexispeter:
please review @kris-lab

modules/kafka/manifests/init.pp
```diff
@@ -0,0 +1,19 @@
+class kafka($version = '0.9.1') {
+
+  require 'apt'
+  require 'php5'
+
+  package { 'librdkafka-dev':
+    ensure   => present,
+    provider => 'apt',
+  }->
```
kris-lab: Just a small note: usually we put this arrow to the new line. It is more readable.

kris-lab:
Added small note in the code.

The rest looks very good!

modules/kafka/manifests/init.pp
```diff
@@ -0,0 +1,19 @@
+class kafka($version = '0.9.1') {
+
+  require 'apt'
+  require 'php5'
```
kris-lab: `require 'build'` is missing?

alexispeter:
please review and merge @kris-lab

kris-lab:
retest this please

PR #1435, opened 2016-08-11
Elasticsearch: stop DEB service after install

kris-lab:


njam:
- Syntax check fails
- Ensure should be "present" and not "latest", otherwise puppet will upgrade the packet for us!
- Check that only 1 "java" process is running in spec
- Will `/etc/init.d/elasticsearch stop` work on Debian 8?

kris-lab:
> Ensure should be "present" and not "latest", otherwise puppet will upgrade the packet for us!

I added latest because the `apt/source` configuration point to elasticsearch `1.3` repository. They have separate one for each version. It means it would not be upgraded to the version `5.0` with current apt/source setup.

> Will /etc/init.d/elasticsearch stop work on Debian 8?

Yes, it works because of systemd controller for that.

kris-lab:
@njam please have a look...

njam:
lgtm

PR #1437, opened 2016-08-12
Elasticsearch: does not stop if package already installed

kris-lab:
Part of https://github.com/cargomedia/puppet-packages/pull/1435 and https://github.com/cargomedia/puppet-packages/pull/1415

For `cargomedia/debian-7-amd64-plain` the spec succeeds, however for `cargomedia/debian-7-amd64-cm` it doesn't! The difference is that `elasticsearch` package is already there so the `stop` command is not executed.

### DIFF
--- a/modules/elasticsearch/manifests/package.pp
+++ b/modules/elasticsearch/manifests/package.pp
@@ -21,7 +21,7 @@ class elasticsearch::package (
 
   exec { 'true && /etc/init.d/elasticsearch stop':
     path         => ['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin'],
-    unless       => 'false',
+    unless       => false,
     require      => Package['elasticsearch'],
     subscribe    => [Apt::Source['elasticsearch'], Package['elasticsearch']],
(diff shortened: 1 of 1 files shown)


kris-lab:
@njam this is working solution for all cases. Please have a look.

kris-lab:
self merging. please add comment @njam

njam:
- `unless => 'false'` is never true, so can be removed?
- I think there should be no different to the previous "unless" though?
- Exec should not require the apt source. Rather the apt package should require the apt source

kris-lab:
Follow up #1438

kris-lab:
> unless => 'false' is never true, so can be removed?

yes, I was thinking it is actually required in puppet: or `unless` or `onlyif` but seems to be not true anymore...

> I think there should be no different to the previous "unless" though?

hmm, previously we checked `/etc/init.d/elasticsearch status` which was failing as this file didn't exist, when the package wasn't there. With installed package the exit code was 0 or 1.

kris-lab:
> Exec should not require the apt source. Rather the apt package should require the apt source

Exec doesn't require apt source but only subscribe for apt source. It require the package to be there.

PR #1439, opened 2016-08-12
Update puppetserver config for version 2.5

njam:
https://docs.puppet.com/puppetserver/latest/release_notes.html#puppet-server-25


njam:
Paired with @kris-lab

Issue #1346, opened 2016-06-01
META: make Janus & services - Jessie compatible


kris-lab:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/645
Part of https://github.com/cargomedia/puppet-packages/issues/1242

- [x] janus
- [x] cm-janus
- [x] `coturn`  (#1384)
- [x] cm::services
- [x] `sk::janus`

cc @ppp0

kris-lab:
@ppp0 as discussed please compile full list of modules needed for migration.

ppp0:
- [ ] ca_certificates
- [ ] database
- [ ] ffmpeg
- [ ] janus
- [ ] lame
- [ ] mysqltuner
- [ ] nodejs
- [ ] systemd
- [ ] helper
- [ ] apt
- [ ] autoprefixer
- [ ] bash
- [ ] bipbip
- [ ] browserify
- [ ] build
- [ ] cgi_fcgi
- [ ] cm
- [ ] cm_janus
- [ ] composer
- [ ] coturn
- [ ] cron
- [ ] daemon
- [ ] elasticsearch
- [ ] environment
- [ ] fluentd
- [ ] foreman
- [ ] gearman
- [ ] git
- [ ] github
- [ ] iptables
- [ ] irqbalance
- [ ] janus
- [ ] kernel
- [ ] lame
- [ ] logrotate
- [ ] memcached
- [ ] mjr_convert
- [ ] mongodb
- [ ] monit
- [ ] mysql
- [ ] mysql_proxy
- [ ] mysqltuner
- [ ] network
- [ ] nginx
- [ ] ntp
- [ ] percona_toolkit
- [ ] phantomjs
- [ ] php5
- [ ] phpunit
- [ ] postfix
- [ ] puppet
- [ ] raid
- [ ] redis
- [ ] rsync
- [ ] rsyslog
- [ ] ruby
- [ ] socket_redis
- [ ] ssh
- [ ] sudo
- [ ] sysctl
- [ ] timezone
- [ ] ufw
- [ ] uglify
- [ ] ulimit
- [ ] vim
- [ ] wowza

kris-lab:
Ok, so we have still 20-30 modules to go to make it compatible? Is it all related sk app?

ppp0:
I went recursively (but manually :( ) through the Puppetfiles

njam:
btw I posted a script to do this in https://github.com/cargomedia/puppet-packages/issues/1242:
[...]

PR #1424, opened 2016-08-15
Memcache for Jessie

kris-lab:
part of #1346


kris-lab:
@njam please review

njam:
If we're not using *daemon* we will not have a monit entry. So on non-wheezy operating systems we will miss the monit entry now.
I think we could just include the monit entry always. Can we use `service memcached start` as a command in the monit entry that works on both sysvinit and systemd?

kris-lab:
> Can we use service memcached start as a command in the monit entry that works on both sysvinit and systemd?

Not sure if I get it right... `service memcached start` works only with `systemd`, however `/etc/init.d/memcached  start` will work with both `systemd` and `sysvinit`. What is your question?

njam:
As discussed we'll try using `daemon` to avoid having different monit entries for sysvinit/systemd.

kris-lab:
@njam have a look please. Please note, I have removed `-d` option, which was breaking in `systemd` and doesn't have effect in `sysvinit`.

njam:
What about `logfile /var/log/memcached.log`, I guess it's needed at least for wheezy?

kris-lab:
Agree, this is bug. However it gets more complicated, because on `systemd` it works out of the box, but on `sysvinit` the have `cat /usr/share/memcached/scripts/start-memcached` wrapepr which creates `PIDFILE` and `LOGFILE`. If we wanna use custom script from `daemon`, we will have to re-use the `start-memcached` for wheezy. wdyt?

njam:
I see!

How about we add a special case (`if`)  for wheezy, which would *not* use daemon, but the default package, and the monit entry? Then we can remove it once not needed anymore.

kris-lab:
@njam have a look....

njam:
lgtm

PR #1406, opened 2016-08-15
Needrestart: manage global configuration

kris-lab:
Part of #1399 

By default `needrestart` add hooks for `dpkg` and `rpm` at `/etc/needrestart/hook.d`. It also configures basic behaviour for detection and auto restart at `/etc/needrestart/needrestart.conf`.

It leads to auto-restart of service by running `apt-get upgrade`. 

I think we should:
- or disable auto-restart or even auto detection
- or remove our custom `/usr/local/bin/needrestart-service` and replace with more fancy `needrestart` hook/configuration scripting

cc @njam @ppp0 

btw: currently it works properly on production

### DIFF
--- a/modules/needrestart/manifests/init.pp
+++ b/modules/needrestart/manifests/init.pp
@@ -1,6 +1,6 @@
 class needrestart (
   $restart_helper_path = '/usr/local/bin/needrestart-service',
-  $auto_apt_invoke = false,
+  $apt_invoke = false,
 ){
 
   package { 'needrestart':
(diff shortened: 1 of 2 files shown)


kris-lab:
@njam please have a look

njam:
Agree with the approach.

I think we could rename the parameter to make its purpose more obvious. "install_apt_hook"? In general this parameter is a bit problematic, because if you set it to "true" *nothing* is done. So if you removed the file before it will not have any impact.
I would rather remove the parameter completely, and add a comment that we overwrite the behaviour to restart all init services by default. wdyt?

kris-lab:
Agree, please have a look @njam

njam:
lgtm

PR #1445, opened 2016-08-17
Jessie: improve php5/opcache extension support

kris-lab:


kris-lab:
@njam please review

njam:
Oh

Did you confirm that this "already loaded" message shows up?
I can see in all other specs we're doing:
```
  describe file('/var/log/php/error.log') do
    its(:content) { should_not match /Warning.*ssh2.*already loaded/ }
  end
```

But if it works with `php --ri` then I think that's better.

If that's confirmed, lgtm

kris-lab:
yes, I confirmed this to show up. The difference is that in `opcache` spec uses `command()`, when in others we use `file()`. Both should work atm.

PR #1449, opened 2016-08-19
Add network::resolvconf

njam:


kris-lab:
I assume this PR is not valid anymore. Will be solved applied to the boxes https://github.com/cargomedia/vagrant-boxes/pull/77

@njam wdyt?

njam:
Agree

PR #1471, opened 2016-08-25
Introduce network::snat

ppp0:
Needed to change the source ip address of outgoing packages (aka source NAT, snat) if we want to fully use Softlayer's Global public IP address as a sender, too

### DIFF
--- a/modules/network/manifests/interface.pp
+++ b/modules/network/manifests/interface.pp
@@ -13,6 +13,7 @@ define network::interface (
   $wpa_ssid     = undef,
   $wpa_psk      = undef,
   $applyconfig  = true,
+  $snat = false,
 ) {
 
   include 'augeas'
@@ -56,6 +57,15 @@ define network::interface (
       path        => ['/bin', '/sbin', '/usr/bin'],
       refreshonly => true,
       subscribe   => Augeas["main-${device}"],
+    }
+  }
+
+  if $snat {
+    $network_interface = regsubst($device, '^([a-z]+\d?)(:?.?)*$', '\1')
+
+    network::snat { "SNAT on ${network_interface}":
+      interface => $network_interface,
+      source_address => $ipaddr,
     }
   }
 
--- a/modules/network/manifests/snat.pp
+++ b/modules/network/manifests/snat.pp
@@ -1,6 +1,6 @@
 define network::snat(
   $interface = undef,
-  $source_address = undef,
+  $source_address,
 ) {
 
   $iface = $interface ? {
--- a/modules/network/manifests/snat.pp
+++ b/modules/network/manifests/snat.pp
@@ -1,7 +1,5 @@
 define network::snat(
   $interface = undef,
-  $from_ip = undef,
-  $to_ip = undef,
   $source_address = undef,
 ) {
 
@@ -10,19 +8,9 @@ define network::snat(
     default => "-o ${interface}",
   }
 
-  $src = $from_ip ? {
-    undef => '',
-    default => "-s ${from_ip}",
-  }
-
-  $dest = $to_ip ? {
-    undef => '',
-    default => "-s ${to_ip}",
-  }
-
   iptables::entry { 'Set up SNAT':
     table => 'nat',
     chain => 'POSTROUTING',
-    rule  => "${iface} ${src} ${dest} -j SNAT --to-source ${source_address}",
+    rule  => "${iface} -j SNAT --to-source ${source_address}",
   }
 }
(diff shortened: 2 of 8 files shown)


ppp0:
@kris-lab please review

To ease reviewing, look at [this commit](https://github.com/cargomedia/puppet-packages/pull/1471/commits/1af834907fd2099f7fd0c9e567cb668bf8b02bd4) separately. There, I just refactor an existing spec test to make use of hiera

modules/network/spec/alias-ip-and-snat/spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'spec_helper'
+
+describe 'network::alias_ip_and_snat' do
+
+  describe command('curl --proxy "" --max-time 1 http://10.10.20.122:1337')  do
+    its(:exit_status) { should be_between(28,52)}
```
ppp0: This is because `curl` under ubuntu returns 52 instead of 28 when timing out

ppp0:
@kris-lab please review

modules/network/manifests/snat.pp
```diff
@@ -0,0 +1,16 @@
+define network::snat(
+  $interface = undef,
+  $source_address = undef,
```
kris-lab: Should it be mandatory? Otherwise what would be the effect of usage?

kris-lab:
Added small comment. Rest lgtm

modules/network/manifests/snat.pp
ppp0: Agree :) - So you don't like `NOOP`s ?

kris-lab: somehow I don't :)

modules/network/spec/alias-ip-and-snat/hiera.json
```diff
@@ -0,0 +1,34 @@
...
+        "xmit_hash_policy": 1
+      },
+      "route_opts": "route add -net 10.10.130.0 netmask 255.255.255.0 gw 10.10.20.128"
+    },
+    "eth1:0": {
+      "method": "static",
+      "ipaddr": "192.168.20.122",
+      "netmask": "255.255.255.0"
+    },
+    "lo:0": {
```
ppp0: This is a "fake" entry needed for the spec tests to run successfully *locally* ( the request would be recognized as to be made on the `lo` interface) - In production, `snat` is used on "real" interfaces

ppp0:
@kris-lab mind reviewing again? I added an argument so that `snat` can be configured out of a hiera tree

modules/network/manifests/interface.pp
```diff
@@ -59,4 +60,13 @@
     }
   }
 
+  if $snat {
+    $network_interface = regsubst($device, '^([a-z]+\d?)(:?.?)*$', '\1')
```
kris-lab: So we do assumption that SNAT interface (virtual) and e.g. Public-IP (regular) run on the same physical interface (e.g. bond0, eth0). I mean, that Virtual-IP/Alias is part of Public-IP interface?

kris-lab: As discussed, this also should work if we introduce multiple NIC (e.g. multiple 10Gbis) with multiple Public-IPs and multiple Global-IPs. @ppp0 agree?

ppp0: Agree. Let's assume you have three interfaces with three different globalIPs. Consider this pseudo hiera tree:

```
[..]
{
  "network::interfaces": {
    "eth0:0": {
      "ipaddress": <globaI IP - alias to eth0>,
      "snat": true,
    },
    "eth1:0": {
      "method": "static",
      "ipaddr": <globaI IP - alias to eth0>,
      "netmask": "255.255.255.0",
      "snat": true,
    },
    "eth2:0": {
      "method": "static",
      "ipaddr": <globaI IP - alias to eth0>,
      "netmask": "255.255.255.0",
      "snat": true
    }
}
[..]
```

This would produce three aliases NICs plus their particular SNAT'ting for traffic leaving through each interface

kris-lab: ok, sgtm

kris-lab:
lgtm

modules/network/manifests/snat.pp
```diff
@@ -0,0 +1,16 @@
...
+  $interface = undef,
+  $source_address,
+) {
+
+  $iface = $interface ? {
+    undef => '',
+    default => "-o ${interface}",
+  }
+
+  iptables::entry { 'Set up SNAT':
```
ppp0: This needs to be unique! @kris-lab cc

PR #1465, opened 2016-08-26
Introduce systemd::critical_units package

tomaszdurka:
- Puppet class for a "critical" target in "systemd"
- Puppet "systemd::unit" should export a virtual resource, which will add a "Wants=" entry to the critical target

### DIFF
--- a/modules/systemd/manifests/critical_unit.pp
+++ b/modules/systemd/manifests/critical_unit.pp
@@ -1,12 +1,12 @@
 define systemd::critical_unit (
-  $serviceName
+  $unitName
 ){
 
   include 'systemd::daemon_reload'
 
-  file { "/etc/systemd/system/critical-units.target.wants/${serviceName}":
+  file { "/etc/systemd/system/critical-units.target.wants/${unitName}":
     ensure  => link,
-    target  => "/etc/systemd/system/${serviceName}",
+    target  => "/etc/systemd/system/${unitName}",
     owner   => '0',
     group   => '0',
     mode    => '0644',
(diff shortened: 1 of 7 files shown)


tomaszdurka:
@njam please review

njam:
- systemd/manifests/critical_unit.pp: $serviceName: Isn't this rather a $unitName (it includes "service" in its name).
 - Same in systemd/manifests/unit.pp
- systemd/manifests/critical_units.pp: It it okay to purge the `.wants/` directory? How will pulsar register the application target?
- systemd/manifests/init.pp: Instead of `class{}` rather use `include` to avoid duplicate definitions if somebody would include the class in hiera?

tomaszdurka:
> - systemd/manifests/critical_unit.pp: $serviceName: Isn't this rather a $unitName (it includes "service" in its name).
>  - Same in systemd/manifests/unit.pp

Sure!

> - systemd/manifests/critical_units.pp: It it okay to purge the `.wants/` directory? How will pulsar register the application target?
Valid point. I added it to actually remove those critical services we dropped. Any suggestions? Should we not purge? We could use `wants` option for that critical unit and let systemd create them? I think it would after reload.

> - systemd/manifests/init.pp: Instead of `class{}` rather use `include` to avoid duplicate definitions if somebody would include the class in hiera?

OK.

njam:
> Should we not purge? We could use `wants` option for that critical unit and let systemd create them? I think it would after reload.

The plan is to use the `.wants/` directory for pulsar, right? Maybe test that before deciding here?
I don't think using the `Wants=` config file option will make a difference. It will be the same like not purging, right?

tomaszdurka:
@njam please final check

njam:
lgtm

PR #1472, opened 2016-08-27
Disable stderr+stdout capturing of systemd daemon unit

njam:
Addressing https://github.com/cargomedia/puppet-cargomedia/issues/1410


njam:
Paired with @kris-lab

PR #1483, opened 2016-09-02
Create proper systemd::unit for critical_units target

tomaszdurka:


tomaszdurka:
@kris-lab or @njam please review

modules/systemd/manifests/unit.pp
```diff
@@ -1,11 +1,15 @@
 define systemd::unit(
   $content,
+  $type = 'service',
+  $critical = true,
 ) {
+  
+  $unitName = "${name}.${type}"
```
kris-lab: maybe better `unit_name`?

kris-lab:
What this new `critical` param exactly does? Is it to avoid new nested unit?

tomaszdurka:
Yes otherwise there was critical-units wants entry for critical-units. On the other hand it didn't make difference... but I would not create that file.

tomaszdurka:
@kris-lab I take that as ok! :)

please confirm!

modules/systemd/manifests/unit.pp
tomaszdurka: Am confused by our conventions. Could and can see some camelcases, but also undercored ones. I see that puppet core has underscores so let's use it.

kris-lab:
yep, that's fine!

PR #1487, opened 2016-09-07
Introduce apt::preferences

ppp0:
Needed for #1486

### DIFF
--- a/modules/apt/manifests/preferences.pp
+++ b/modules/apt/manifests/preferences.pp
@@ -1,0 +1,18 @@
+define apt::preferences (
+  $package = $title,
+  $pin,
+  $pin_priority = 1001,
+) {
+
+  require 'apt'
+  include 'apt::update'
+
+  file { "/etc/apt/preferences.d/${package}.pref":
+    ensure  => file,
+    content => template("${module_name}/preferences.erb"),
+    owner   => 0,
+    group   => 0,
+    mode    => '0644',
+    notify  => Exec['apt_update'],
+  }
+}
--- a/modules/apt/spec/preferences/manifest.pp
+++ b/modules/apt/spec/preferences/manifest.pp
@@ -1,0 +1,16 @@
+node default {
+
+  $version = 'version 8:6.8.9.9-5+deb8u2'
+
+  apt::preferences { 'imagemagick-common':
+    pin => $version,
+  }
+
+  apt::preferences { 'libmagickcore-6.q16-2':
+    pin => $version,
+  }
+
+  apt::preferences { 'libmagickwand-6.q16-2':
+    pin => $version,
+  }
+}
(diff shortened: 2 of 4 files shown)


ppp0:
retest this please

modules/apt/spec/preferences/spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    its(:content) { should match // }
+  end
+
+  describe file('/etc/apt/preferences.d/libmagickwand-6.q16-2.pref') do
+    its(:content) { should match // }
+  end
+
+  describe command('sudo apt-get -y install imagemagick') do
+    its(:exit_status) { should eq 0 }
+    its(:stdout) { should match /Unpacking imagemagick-common \(8:6\.8\.9\.9-5\+deb8u2\)/}
```
ppp0: I don't like this spec test. Packages will evolve and it won't be so easy to match that version

Any ideas? @kris-lab @njam ?

modules/apt/manifests/preferences.pp
```diff
@@ -0,0 +1,18 @@
+define apt::preferences (
+  $package = $title,
+  $pin,
```
kris-lab: maybe we should call it `pin_version` to clarify what is that?

modules/apt/spec/preferences/manifest.pp
```diff
@@ -0,0 +1,16 @@
+node default {
+
+  $version = 'version 8:6.8.9.9-5+deb8u2'
+
```
ppp0: For convenience, I could imagine introducing a class `apt::preferences::imagemagick_stable` or similar which wraps the follwing three statements. @njam @kris-lab agree?

modules/apt/spec/preferences/spec.rb
kris-lab: I think we should not test like that. This version will be removed from repo rather sooner than later. We know it works, that the mechanism works fine yes?

kris-lab:
In general let's verify only that module write properly all apt/preferences files. Testing specific version with specs might be tricky as we use apt repos (version of packages will change in time), wdyt?

modules/apt/spec/preferences/manifest.pp
kris-lab: The only thing we can achieve is to install specific version or with apt or from url (DEB). Then we should apply this module/requirement. This is because the package with such version will not stay forever in the apt imo.

njam:
Agree, I would just check that the APT settings are stored correctly. Installing a specific version of a package is likely to fail soon.

What is currently asserted in the spec test could be moved to imagemagick's spec test?

kris-lab:
Taking over. @njam please review

njam:
lgtm

PR #1492, opened 2016-09-08
systemd: Use configuration files instead of symlink for units dependencies

tomaszdurka:
I've played a bit more with systemd and found that some of the links from wants/requires directories (those manually created) are removed in chains - systemd thinks they are not needed anymore.

I will rework module to work on `/etc/systemd/system/{$service}.d/foo.conf` files.

### DIFF
--- a/modules/systemd/manifests/critical_unit.pp
+++ b/modules/systemd/manifests/critical_unit.pp
@@ -2,11 +2,13 @@ define systemd::critical_unit {
 
   include 'systemd::daemon_reload'
   
-  $wants = $name
+  $unit_configuration = {
+    'wants' => $name,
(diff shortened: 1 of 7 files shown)


tomaszdurka:
@kris-lab or @ppp0 please review

njam:
lgtm

Maybe call `systemd/templates/conf` rather `systemd/templates/unit.erb`?

tomaszdurka:
Test passed, there was problem with CI cleaning the vagrant env. Is that actually a problem for @cargomedia/devops ?

tomaszdurka:
@njam please review or delegate

njam:
lgtm

PR #1499, opened 2016-09-12
Rework NAT/SNAT handling

ppp0:


ppp0:
@kris-lab please review

I've done a rework of nat/snat. If a `source_address` is provided, the the target will be `SNAT` and the source address will be changed to the given address. If not, all stays as it was (using target `MASQUERADE`).

New: in any case, the rule to accept incoming requests are NAT'ted if they originate from a address "behind".

kris-lab:
I think this is good and safe for us. As far as I can see the internal difference is very little between `MASQUERADE` and `SNAT`:
>  http://unix.stackexchange.com/a/21968
The SNAT target requires you to give it an IP address to apply to all the outgoing packets. The MASQUERADE target lets you give it an interface, and whatever address is on that interface is the address that is applied to all the outgoing packets. In addition, with SNAT, the kernel's connection tracking keeps track of all the connections when the interface is taken down and brought back up; the same is not true for the MASQUERADE target.

However I am wondering if keeping all connection in kernel works fine for us? I think yes, wdyt?

kris-lab:
code lgtm

ppp0:
@kris-lab please review

as discussed: let's always use SNAT here as we don't have to deal with dynamic addresses (DHCP) in our LAN

kris-lab:
OK, but then we will have to set global IP for each datacenter in `puppet-cargomedia`, yes?

ppp0:
Can't we set the source IP to the machine offering the NAT gateway? Global IP where needed

PR #1497, opened 2016-09-13
Move monit entry for daemon into sysvinit version only

tomaszdurka:


tomaszdurka:
@ppp0 pleae review

ppp0:
makes sense, lgtm

kris-lab:
@tomaszdurka what is the status here?

tomaszdurka:
Oh, mergable - except build failed.
Will adjust and ask for review/

tomaszdurka:
Checking status.

tomaszdurka:
Retest this please.

ppp0:
lgtm

njam:
It looks like `monit.systemd.erb` is unused now and can be removed.

Also monit's `conf.d` is *not* purged with puppet (because we put entries from the application there). Either we should clean up production, or check if we can purge the directory now. Second one is probably better.

@tomaszdurka can you open a followup?

tomaszdurka:
Done in #1523.

PR #1507, opened 2016-09-15
Update composer

fauvel:


fauvel:
@njam please review

njam:
@tomaszdurka can you look into this?

njam:
@kris-lab @ppp0 could you help Sébastien here?

kris-lab:
Some reason new composer doesn't create proper permission for binary file in the project.
```
$ ./usr/local/lib/jsonlint/bin/jsonlint  --help
bash: ./usr/local/lib/jsonlint/bin/jsonlint: No such file or directory
```
By changing permission it seems to be working:
```
$ chmod +x /usr/local/lib/jsonlint/bin/jsonlint 
$ /usr/local/lib/jsonlint/bin/jsonlint  --help
Usage: jsonlint file [options]

Options:
  -q, --quiet     Cause jsonlint to be quiet when no errors are found
  -h, --help      Show this message

```
Will investigate deeper, however any input very welcome, wdyt @njam ?

tomaszdurka:
So what are the permissions for the created symlink file?

kris-lab:
@tomaszdurka what symlink do you have in mind?
```
# ls -lpa /usr/local/lib/jsonlint/bin/jsonlint
-rw-r--r-- 1 root root 2588 Sep 16 09:58 /usr/local/lib/jsonlint/bin/jsonlint
```

njam:
Strange… I'm wondering if the original file should be executable or not.
If so, then we can possibly just use a different package with a different binary for this spec?

kris-lab:
Well, I agree. I thought that according to the [documentation](https://getcomposer.org/doc/articles/vendor-binaries.md#how-is-it-defined-) and configuration of [jsonlint](https://github.com/Seldaek/jsonlint/blob/master/composer.json#L20) the `+x` flag should be set by composer, isn't it?

njam:
Agree, also the file is executable in the source repo:
https://github.com/Seldaek/jsonlint/blob/master/bin/jsonlint

tomaszdurka:
@kris-lab Sorry got confused. I thought these executables are actually symlinks to original (repo-based) files.

fauvel:
@kris-lab @tomaszdurka are you taking over here?

tomaszdurka:
I'm not sure what shall be done here? These are internal composer mechanisms. Is there a non-hacky solution?

ppp0:
I'd say a directory with name `bin/` shall contain (mostly) executables... If composer is not able to set the `+x` flag we should give him a help. Does this sound hacky? I spent already too much time imo to find out about the inner workings of Composer (it was about spitting all output to STDERR iirc) and I would not dare touch too much there.

kris-lab:
@ppp0 could you take over from here?

ppp0:
Yes. IMO we can safely merge this one as it is. Reasons:
* Spec test failing on an external repo (seld/jsonlint) - not a composer issue?
* Spec failing **on Wheezy only** not on Jessie
* All the rest lloks good on both Wheezy and Jessie

I would say supporting Wheezy is less and less (may be even not anymore) a concern.

@njam you have the very last word

njam:
Agree, spec succeeds on Debian 8/PHP 5.6.
I think we can safely merge this, as we'll remove Debian 7/PHP 5.4 support hopefully next week.

PR #1502, opened 2016-09-21
Refactor php-rdkafka module

tomaszdurka:
Added: https://github.com/cargomedia/puppet-packages/pull/1433
Once wheezy support for "php" is removed.

- Move everything to `php::extension::rdkafka`

cc @alexispeter


tomaszdurka:
@ppp0 please review

tomaszdurka:
@alexispeter wanna review?

alexispeter:
Not sure what effect removing the puppetfile and metadata.json has. The rest looks good.

njam:
Does this succeed on Debian 7? iirc we extracted this into a separate module to declare Debian 8-only compatibility. We could also remove Debian 7-support for the "php" module now!

tomaszdurka:
Yes I thought it won't have support for Debian-7 and we should rather drop support in php5, wdyt?

njam:
agree

ppp0:
lgtm

tomaszdurka:
so let's not wait on the build, I tested extension locally and worked fine

tomaszdurka:
I need a /nod for that :)

njam:
build is green, lgtm

PR #1474, opened 2016-09-23
Verify if "sysctl::entry" is applied initially, after change and on reboot

kris-lab:


kris-lab:
@njam @ppp0 I have added new spec and some refactoring, but I cannot reproduce the problem we experience on production. Entries have been always properly applied. wdyt?

kris-lab:
Sorry guys for confusing you. I have double checked and I can reproduce this problem. Missing `.conf` is the reason for not applying the config after reboot.

@njam @ppp0 please review

njam:
👍 lgtm

njam:
I think old files are still on production.
Maybe delete them everywhere to avoid confusion?

kris-lab:
Will run to cleanup:
```
find /etc/sysctl.d/ -type f ! -name "*.*" -delete
```

kris-lab:
Done!

PR #1514, opened 2016-09-24
cm::vhost: enable gzip for AJAX requests by default

kris-lab:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1518

### DIFF
--- a/modules/nginx/templates/conf.d/nginx.conf.erb
+++ b/modules/nginx/templates/conf.d/nginx.conf.erb
@@ -16,7 +16,11 @@ http {
 
   sendfile                  <%= scope.lookupvar('nginx::params::nx_sendfile')%>;
   <% if scope.lookupvar('nginx::params::nx_gzip') == 'on'%>
-  gzip                      on;
+  gzip                        on;
+  gzip_proxied                any;
+  gzip_http_version           1.0;
+  gzip_min_length             1000;
+  gzip_types application/json application/x-javascript text/css text/plain application/xml image/svg+xml;
   <% end %>
   client_max_body_size      <%= @client_max_body_size %>;
   keepalive_timeout         <%= @keepalive_timeout %>;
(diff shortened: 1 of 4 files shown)


kris-lab:
@njam please have a look. Confirmed to be working with XHR requests. I have added all types which are currently used in other vhost/locations for cm services. However the `image/svg+xml;  text/css text/plain` could be most likely dropped there (in our case).

njam:
Agree on the value.

I would suggest to apply this value everywhere where we enable GZIP. How about we use this gzip configuration as a default (in the `http` configuration), and remove it from all the `location`? I think we *always* want this config by default, and if not, we can override it.

(same goes for some SSL config like `ssl_session_timeout`, `ssl_session_cache`, `ssl_prefer_server_ciphers`)

kris-lab:
> (same goes for some SSL config like ssl_session_timeout, ssl_session_cache, ssl_prefer_server_ciphers)

however this is configured properly atm, agree?

kris-lab:
@njam please review

kris-lab:
@njam please re-review

modules/nginx/templates/conf.d/nginx.conf.erb
```diff
@@ -16,7 +16,11 @@ http {
 
   sendfile                  <%= scope.lookupvar('nginx::params::nx_sendfile')%>;
   <% if scope.lookupvar('nginx::params::nx_gzip') == 'on'%>
-  gzip                      on;
+  gzip                        on;
+  gzip_proxied                any;
+  gzip_http_version           1.0;
+  gzip_min_length             1000;
+  gzip_types application/json application/x-javascript text/css text/plain application/xml image/svg+xml;
```
njam: Let's add:
- text/html: it's enabled by default, so currently active for some locations. We should keep it.
- application/javascript: It's the proper type for JS, we don't currently use it, but let's add it for compatibility.

kris-lab:
@njam please review

kris-lab:
@njam could also confirm it works as expected?

njam:
lgtm!

PR #1515, opened 2016-09-26
mysqld - Increase limit of max connections

ppp0:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1541

### DIFF
--- a/modules/mysql/templates/my.cnf
+++ b/modules/mysql/templates/my.cnf
@@ -26,8 +26,8 @@ skip-name-resolve
 key_buffer = 8192M
 max_allowed_packet = 16M
 thread_stack = 512K
-thread_cache_size = 1500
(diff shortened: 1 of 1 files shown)


ppp0:
@kris-lab please review.

This needs careful merging as it will restart the mysql server!.. May be it's a good idea to halt the application, gracefully stop db master, apply to the slaves first and then to the master, restart app

kris-lab:
I am wondering if `max_connections = 1920` is enough, but let's see on production.

ppp0:
I would not raise too much in one go, this for following reasons:
* 1500 was enough until before migrating (It was mysql 5.5 on Wheezy)
* one connection is one more thread incl. its overhead and costs

Let's see how far we get with these settings

PR #1503, opened 2016-09-26
Service on template does not match variable

tomaszdurka:
Variable from template:
- https://github.com/cargomedia/puppet-packages/blob/master/modules/raid/templates/linux_md/monit.systemd.erb#L1

doesn't match one from manifest: 
- https://github.com/cargomedia/puppet-packages/blob/master/modules/raid/manifests/linux_md.pp#L6

### DIFF
--- a/modules/raid/manifests/linux_md.pp
+++ b/modules/raid/manifests/linux_md.pp
@@ -3,9 +3,9 @@ class raid::linux_md {
   require 'apt'
 
   if ($::facts['lsbdistcodename'] == 'vivid') {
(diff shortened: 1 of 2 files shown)


tomaszdurka:
@ppp0 please review

ppp0:
the correct spelling is `mdadm` (adm being admin), so may be change this where the variable gets set?

ppp0:
@tomaszdurka fyi

tomaszdurka:
lgtm

PR #1520, opened 2016-09-27
Reduce defaults for mysql

ppp0:


ppp0:
Further lowering the requirements, self-merging
@kris-lab @njam fyi

PR #1508, opened 2016-09-27
Make iptables::entry persistent

ppp0:


ppp0:
@kris-lab I provide two versions here with identical functionality
* `ufw::nat`
* `network::nat`

I like how we use `ufw` and streamline how we touch `iptables` (plus the persistence included)

But, in a discussion with @njam some downsides came up:

* ufw needs to be installed where we want NAT
* Servers are already behind a HW firewall

Alternatively, there is https://packages.debian.org/jessie/iptables-persistent but I would not introduce another tool again that manipulates `iptables`

what we need to make sure is, that were we install `ufw::nat`, we either open the private interface to all traffic (default for `ufw` is to drop everything) and/or configure ufw to allow all applications that we have installed.

Good practice would anyway be that apps come with their `/etc/ufw/application.d` file - or that we provide a firewall rule

please have a look

kris-lab:
Both approaches looks fine with me. Finally, both have the same layout for network rules, so I would choose the one which introduce less overhead. I believe UFW will introduce more complexity at the moment, wdyt?

I fully agree that we should ship apps with firewall rules in the future (it should be a standard package: bipbip/monitoring, firewall/security, monit/alerting etc..).

njam:
One thought about UFW rules that we discussed with Philipp yesterday: for our production setup - which consists of multiple servers that communicate over a private network - we need to make sure that the private interface is open for most/all traffic. E.g. a memcache server needs to have port 11211 open on the private interface.

ppp0:
Yes, quoting myself
>what we need to make sure is, that were we install ufw::nat, we either open the private interface to all traffic (default for ufw is to drop everything) and/or configure ufw to allow all applications that we have installed.

ppp0:
The magic command is `ufw allow in on eth1 to any` where `eth1` would be changed to whatever is the private nic

ppp0:
As discussed w/ @kris-lab let's keep all shut by default (Good security practice) and open in puppet cargomedia for datacenters behind a hardware fw

ppp0:
Changes:

* as discussed, new `network::nat` class that will use `ufw` and thus **provide persistence**
* New default for `ufw::init` is to allow all traffic originating from well-known private addresses
* Remove unused `iptables` toolset
* Remove `ufw::rule` to allow `22/ssh` in spec tests (default now is open on private networks)

@kris-lab please have a look

ppp0:
retest this please

kris-lab:
> Remove ufw::rule to allow 22/ssh in spec tests (default now is open on private networks)

What about servers like janus[X] or bulldog[X] where we do have SSH access via public interface. Currently I see `@ufw::rule { 'ssh': app_or_port => 22 }` removed everywhere. Is that correct cuz opening only private network SSH access could not be enough?

kris-lab:
(btw, My question is not fully related to quoted statement, but you know what I meant)

modules/network/manifests/nat.pp
```diff
@@ -4,27 +4,21 @@
...
-  iptables::entry { 'Allow inbound traffic for established connection':
-    chain => 'FORWARD',
-    rule  => "-i ${ifname_public} -o ${ifname_private} -m state --state RELATED,ESTABLISHED -j ACCEPT",
-  }
-  ->
-
-  iptables::entry { 'Allow outbound traffic':
-    chain => 'FORWARD',
-    rule  => "-i ${ifname_private} -o ${ifname_public} -j ACCEPT",
+  file {
```
ppp0: agree, conceptually better to call a define that touches the module's config

ppp0:
>>Remove ufw::rule to allow 22/ssh in spec tests (default now is open on private networks)

>What about servers like janus[X] or bulldog[X] where we do have SSH access via public interface. Currently I see @ufw::rule { 'ssh': app_or_port => 22 } removed everywhere. Is that correct cuz opening only private network SSH access could not be enough?

* Remove ufw::rule **only** relates to spec tests where we don't need anymore to explicitly open 22/ssh to run the tests
* Wherever we open a port on a public interface, nothing really changes, ie we need to open 22/tcp on bulldogs, otherwise we'd be cut off

ppp0:
Changes
* Introduce resource for rules/before (rules/after could be easily added if needed)
* Fix test (`ufw::default`)
* Fix service activation/running ordering. I found out that on Debian-8 (not on Ubuntu?), `ufw` was "enabled" but not "active" (ie "running" in systemd terminology)

@kris-lab please have a big look

ppp0:
modifying polipo made things even worse :/

kris-lab:
What is the status here?

ppp0:
status is that builds fail because of network issues (package download et al.) but code is ready, did you review it?

modules/network/templates/snat.rules.erb
```diff
@@ -0,0 +1,14 @@
...
+*nat
+:POSTROUTING ACCEPT [0:0]
+# Modify outgoing traffic's source IP
+-A POSTROUTING -o <%= @ifname_public %> -j SNAT --to-source <%= @to_source %>
+COMMIT
+
+*filter
+:puppet-nat - [0:0]
+# Allow outgoing traffic
+-A puppet-nat -i <%= @ifname_private %> -o <%= @ifname_public %> -j ACCEPT
```
kris-lab: `-state RELATED,ESTABLISHED` not needed anymore?

kris-lab:
- ./modules/network/Puppetfile: remove `iptables`
- `/etc/ufw/before.d` I see we purge that directory and then subscribe within `ufw::service`. Will it run every puppet round?

ppp0:
>/etc/ufw/before.d I see we purge that directory and then subscribe within ufw::service. Will it run every puppet round?

We do the same in https://github.com/cargomedia/puppet-packages/blob/master/modules/nginx/manifests/service.pp#L6 and https://github.com/cargomedia/puppet-packages/blob/master/modules/nginx/manifests/config.pp#L71 actually.. I will test, though

kris-lab:
Yes, please test this, but I agree it should work. I misread this the workflow. I thought we subscribe to the directory and based on change we do reload. However we reload only if all catenated files output has changes (within dir), which is good!

ppp0:
looks good! Not reapplied as expected. @kris-lab please final look

kris-lab:
code looks good to me. 

What is the deployment procedure?
- all puppet agents off and apply manually on some testing hosts: janus, www/snat based?
- then on all others if test passed?

ppp0:
We have two hosts that include `network::nat` (with args in hiera). 
* [www8](https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/files/hiera-data/datacenter/softlayer/vlan1850-metal/www8.json#L20)
* [stream3](https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/files/hiera-data/datacenter/softlayer/vlan1850-metal/stream3.json#L20)

There will be a short downtime - may be even none - when we flush and re-apply the iptables rules.
* Stop puppet
* `systemctl stop ufw` (will invoke `ufw_stop` in `/lib/ufw/ufw-init-functions` which will clear all ufw managed chains and rules
* Exec this to completely flush `iptables`:
  
 ```
iptables -P INPUT ACCEPT
iptables -P FORWARD ACCEPT
iptables -P OUTPUT ACCEPT
iptables -t nat -F
iptables -t mangle -F
iptables -F
iptables -X
   ````
* `puppet agent -t`
* restart puppet

@kris-lab stop the app while doing this? not sure...

kris-lab:
I would schedule downtime, it should be short however will affect all connected clients.

ppp0:
retest (with squid instead of polipo)

ppp0:
retest this please

kris-lab:
@ppp0 let's pair and merge this asap

ppp0:
We gave a first try at migration and came across several problems (reason for reverting). Apparently, `cat`ing "before rules" [as we tried](https://github.com/ppp0/puppet-packages/blob/8f993b05a2758e49230e866b0dd862be266abf19/modules/ufw/manifests/service.pp#L4) is not the proper way to go: `ufw` gets confused. The proper way to test this is to completely flush `iptables` before applying these rules:

```
iptables -P INPUT ACCEPT
iptables -P FORWARD ACCEPT
iptables -P OUTPUT ACCEPT
iptables -t nat -F
iptables -t mangle -F
iptables -F
iptables -X
``` 

Also, rules that used to open the firewall towards the private network, such as 
* https://github.com/cargomedia/puppet-cargomedia/blob/88472c9b4ff64ae0380d05768358f04918ec592d/modules/cargomedia/manifests/mongodb/router.pp#L23
* https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/manifests/mongodb/role.pp#L24

PR #1524, opened 2016-09-29
Remove Wheezy Support

ppp0:
Part of https://github.com/cargomedia/roadmap/issues/95

PR #1528, opened 2016-10-03
Puppet runs failing

ppp0:
Regression introduced in #1524


ppp0:
@kris-lab fyi - we forgot to remove an obsolete arg

PR #1530, opened 2016-10-03
Debug memcached unit

ppp0:
Regression introduced in #1524


ppp0:
@kris-lab fyi

The memcached user has changed to memcached (from 'nobody') which is correct. But the unit still executes as root!

PR #1533, opened 2016-10-05
Make sure "[/etc/monit/conf.d/alert" not removed every run

kris-lab:
```
Notice: /Stage[main]/Monit/File[/etc/monit/conf.d/alert]/ensure: removed
Notice: /Stage[main]/Monit/Exec[/etc/monit/conf.d/alert]/returns: executed successfully
```

### DIFF
--- a/modules/monit/manifests/init.pp
+++ b/modules/monit/manifests/init.pp
@@ -32,10 +32,10 @@ class monit ($emailTo = 'root@localhost', $emailFrom = undef, $allowedHosts = []
   ->
 
   file { '/etc/monit/conf.d':
-    ensure  => directory,
-    group   => '0',
-    owner   => '0',
-    mode    => '0755',
+    ensure => directory,
+    group  => '0',
+    owner  => '0',
+    mode   => '0755',
     purge   => true,
     recurse => true,
   }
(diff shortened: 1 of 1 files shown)


kris-lab:
@ppp0 @njam since we purge `/etc/monit/conf.d` we cannot easily let puppet know that some files/symlinks should be persistent and the output of `exec` with flag `creates` should be considered like normal file. 

Instead of using `exec` I simply create this file with initial content as workaround. This happens only once during installation. It is sensitive for content change (which is good) + it doesn't conflict with normal operation of `monit-alert <profile>` procedure.

I am gonna merge this, due to many alerts.

let me know wdyt?

kris-lab:
I have simplified a little bit. Currently it:
- creates symlink if not existing and point to default profile
- does not create if symlink exists and points to not default profile

njam:
The idea was to remove this functionality completely @ppp0 @tomaszdurka right?

kris-lab:
The idea was good of course, but not done fully. We cannot have 50 open tickets in copperegg so this is workaround for the moment.

njam:
I understand, makes sense. :)

PR #1534, opened 2016-10-05
Fix failing db backup script

ppp0:


ppp0:
@kris-lab please review

kris-lab:
in general looks good. What about `critical-units.target` instead of monit, will it alert?

ppp0:
It will, after 10min. But the critical operation is expected to last less. Critical is when we stop the mysql server on db9 to copy it's /var/lib to a temp directory. This takes some minutes due to its size. After that, we restart and the critical period is over. Of course, the backup proper afterwards takes hours to complete

ppp0:
But I get you: shall we implement some silencing of `critical-units.target` ?

kris-lab:
Let's keep it like is now. If this 10min is too less, then let's find better solution. Agree?

ppp0:
fully agree

Issue #1532, opened 2016-10-04
Remove sysvinit stuff


njam:
There are some usages of init scripts around that can/should be removed/changed.

@ppp0 @kris-lab can you check it?

ppp0:
The only place I see which could need a refurbishment in `mysql::service`
The rest is either to be deleted by #1498 or by refactoring our log rotation
Where do you see other usages that need intervention?

njam:
- backup/templates/agent/rdiff/create.sh
- mysql/manifests/service.pp
- mysql/templates/replication-check.sh
- puppetserver/manifests/puppetdb/certs.pp

wdyt?

ppp0:
closing as completed. Reopen if you disagree

PR #1535, opened 2016-10-06
Remove sysvinit from puppetserver

kris-lab:
Part of https://github.com/cargomedia/puppet-packages/issues/1532


kris-lab:
@ppp0 wdyt?

ppp0:
I think it's a good idea

PR #1540, opened 2016-10-06
Add arg critical to daemon

ppp0:
The idea is not to declare every daemon we create as 'critical'


ppp0:
@kris-lab please review

kris-lab:
lgtm, just to double confirm, it should not change anything on production. Currently there is ` $critical = true` for all running services, agree?

ppp0:
Yes, correct!

Issue #1430, opened 2016-08-04
Disable log for mongodb


ppp0:
Similar to https://github.com/cargomedia/puppet-packages/issues/1377

In systemd all daemons should just dump their output to stdout stderr
journald will take care of storing/rotating

No more log files all over the place, log rotation etc.

Solves https://github.com/cargomedia/puppet-packages/pull/1421

ppp0:
Keep wheezy compatibilty until phased out

kris-lab:
sgtm

kris-lab:
Temporary closing this ticket for better visibility. Will be done according to finding in https://github.com/cargomedia/puppet-cargomedia/issues/1411#issuecomment-258804467

Issue #1377, opened 2016-06-24
Disable log file of janus


njam:
After https://github.com/cargomedia/puppet-packages/issues/1346

With systemd janus logs directly to journald.
So we should disable logging to a separate file, and remove the fluentd source (because we have it from syslog).

cc @ppp0

kris-lab:
Temporary closing this ticket for better visibility. Will be done according to finding in https://github.com/cargomedia/puppet-cargomedia/issues/1411#issuecomment-258804467

PR #1541, opened 2016-10-06
Configurable verbosity of daemons

ppp0:
Systemd's default is to capture STDOUT/STDERR of services and journaling them. This seems to be muted/overridden in our current setup.

This would conflict with
* https://github.com/cargomedia/puppet-packages/issues/1430
* https://github.com/cargomedia/puppet-packages/issues/1377

and with the move away from log files in general

### DIFF
--- a/modules/daemon/templates/systemd.service.erb
+++ b/modules/daemon/templates/systemd.service.erb
@@ -52,8 +52,6 @@ RuntimeDirectoryMode=<%= @runtime_directory_mode %>
 <% @env.each do |name, value| -%>
 Environment="<%= name %>=<%= value %>"
 <% end -%>
-StandardOutput=null
-StandardError=null
 
 [Install]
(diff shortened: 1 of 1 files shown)


ppp0:
@kris-lab please review

kris-lab:
Well, this was introduced because of logs flood in loggly. See https://github.com/cargomedia/puppet-cargomedia/issues/1410

Removing this will revive the problem for the moment.

ppp0:
I understand.. So what to do with our logging? I still think it is good to centralise it.

njam:
There's a bigger picture discussion in https://github.com/cargomedia/puppet-cargomedia/issues/1411. I would suggest to first decide what the goal is there, do some experiments on how it can be achieved, and then apply it for all services.

kris-lab:
As discussed closing for the moment.

PR #1023, opened 2015-12-03
Upgrade satis version

njam:

Issue #1005, opened 2015-11-23
Spec fails: satis::repo


njam:

njam:
This error on CI:
```
Could not authenticate against github.com
```
full output:
```
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: Note: checking out '3dc787c16f1082784808e4a09d0e9e2fb2661afc'.
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: 
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: You are in 'detached HEAD' state. You can look around, make experimental
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: changes and commit them, and you can discard any commits you make in this
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: state without impacting any branches by performing another checkout.
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: 
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: If you want to create a new branch to retain commits you create, you may
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: do so (now or later) by using -b with the checkout command again. Example:
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: 
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns:   git checkout -b new_branch_name
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: 
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: HEAD is now at 3dc787c... Update deps
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: Loading composer repositories with package information
      Notice: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: Installing dependencies from lock file
[...]

PR #1328, opened 2016-10-06
Run satis as a daemon/service instead of cron job

ppp0:
As discussed, we gain more control about handling error conditions if we run `satis` as a daemon
E.g. we can retry downloads a couple of times before throwing alerts

Continuation of #1312 #1327


ppp0:
@kris-lab please review

The idea is to have a long-running daemon instead of a cron job which mails us in case of failures. Failures are, as we have seen, often due to github issues leading to network exceptions

ppp0:
Sorry @kris-lab please consider only commit https://github.com/cargomedia/puppet-packages/pull/1328/commits/5db46c049aa9cd934df09af0f3dacea7cf6da429

njam:
If upgrading the repo fails for a longer time, it interrupts developer workflows.
There fore I think we should retry a "couple of times", but then fail and inform a sysadmin.
Or maybe even detect some special errors (like github down) and ignore only those?

ppp0:
retest this please

ppp0:
OK, so now the "daemon" is modify to ignore TransportException AND github.com - All others will result in a failure. A failure will dump the content of the error (temp) file to the journal and the "daemon" will be restarted by systemd as per our default (via `daemon` resource)

@kris-lab please review

kris-lab:
Code looks good!

kris-lab:
retest this please

kris-lab:
@ppp0  we get this error on build:
````
 Error: Command exceeded timeout
      Error: /Stage[main]/Satis/Composer::Project[composer/satis]/Exec[upgrade composer/satis]/returns: change from notrun to 0 failed: Command exceeded timeout
```
do you know why?

ppp0:
I tried now several times both on my local VM and on ci5's (where the build failed) and is runs through... Really weird. Might be solved by https://github.com/cargomedia/puppet-packages/issues/885 ? I don't know. It looks like it's failing consistently on that `composer upgrade` step (as always, no comments about composer)

ppp0:
Let's try with an updated satis version.. (see https://github.com/cargomedia/puppet-packages/pull/1328/commits/0749ac83047a45413de14a7e802865390c8085c6) - It helped on https://github.com/cargomedia/puppet-packages/issues/1005 and https://github.com/cargomedia/puppet-packages/pull/1023

ppp0:
@kris-lab I can't believe it, but upgrading satis has helped. please review

kris-lab:
ok! lgtm

PR #1543, opened 2016-10-10
Update net-ssh

ppp0:


ppp0:
Same as in https://github.com/cargomedia/pulsar-conf-cargomedia/pull/356

@kris-lab please review

njam:
Copying from https://github.com/cargomedia/pulsar-conf-cargomedia/pull/356#issuecomment-252825033:
> Shouldn't we always use pessimistic version constraints, at least `~> x.y`, if not `~> x.y.z`?
Not using any version constraint means that whenever a breaking change is released the software in question might break too.

PR #1542, opened 2016-10-10
Managing ucf.conf to force confold on package install/upgrade

ppp0:
We have a new kind of failure during unattended upgrade

```
root@xxx:/home/admin$ apt-get upgrade
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Calculating upgrade... Done
The following packages will be upgraded:
  php-pear php5-cli php5-common php5-curl php5-dev php5-fpm php5-gd php5-intl php5-mcrypt php5-mysql php5-readline
11 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

screen shot 2016-10-10 at 12 34 54

Introducing `/etc/apt/apt.conf.d/33-force-conf` with
```
Dpkg::Options {
   "--force-confdef";
   "--force-confold";
};
```

~~should fix it~~ UPDATE: It does not

### DIFF
--- a/modules/ucf/manifests/init.pp
+++ b/modules/ucf/manifests/init.pp
@@ -11,6 +11,8 @@ class ucf {
     before  => Package['ucf'],
   }
 
-  package { 'ucf': }
-
+  package { 'ucf':
+    ensure   => present,
+    provider => 'apt',
+  }
 }
--- a/modules/ucf/manifests/init.pp
+++ b/modules/ucf/manifests/init.pp
@@ -1,0 +1,16 @@
+class ucf {
+
+  require 'apt'
+
+  file { '/etc/ucf.conf':
+    ensure  => file,
+    content => template("${module_name}/ucf.conf"),
+    owner   => '0',
+    group   => '0',
+    mode    => '0644',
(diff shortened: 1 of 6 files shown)


ppp0:
@kris-lab please review

modules/ucf/manifests/init.pp
```diff
@@ -0,0 +1,16 @@
...
+  file { '/etc/ucf.conf':
+    ensure  => file,
+    content => template("${module_name}/ucf.conf"),
+    owner   => '0',
+    group   => '0',
+    mode    => '0644',
+    before  => Package['ucf'],
+  }
+
+  package { 'ucf': }
```
kris-lab: Maybe also?
```
    ensure   => present,
    provider => 'apt',
```

kris-lab:
@ppp0 added small comment. The rest looks very good!

modules/ucf/manifests/init.pp
ppp0: Thanks!

ppp0:
merged @kris-lab fyi

PR #1545, opened 2016-10-11
Added network route update spec

kris-lab:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1472

### DIFF
--- a/modules/network/spec/route-change/1-configure.pp
+++ b/modules/network/spec/route-change/1-configure.pp
@@ -1,0 +1,9 @@
+node default {
+
+  network::interface { 'eth1':
+    method       => 'static',
+    ipaddr       => '10.10.20.122',
+    netmask      => '255.255.255.0',
+    route_opts   => 'route add -net 10.10.130.0 netmask 255.255.255.0 gw 10.10.20.128',
+  }
+}
--- a/modules/network/spec/route-change/2-reconfigure.pp
+++ b/modules/network/spec/route-change/2-reconfigure.pp
@@ -1,0 +1,9 @@
+node default {
+
+  network::interface { 'eth1':
+    method       => 'static',
+    ipaddr       => '10.10.20.122',
+    netmask      => '255.255.255.0',
+    route_opts   => 'route add -net 10.10.134.0 netmask 255.255.255.128 gw 10.10.20.128; route add -net 10.10.138.0 netmask 255.255.255.192 gw 10.10.20.138',
+  }
+}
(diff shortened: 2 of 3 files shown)


kris-lab:
@ppp0 I have added spec and done manual test for bug reported here https://github.com/cargomedia/puppet-cargomedia/issues/1472 however I cannot reproduce this problem. Could you have a look what I missed there?

kris-lab:
Btw `/etc/network//interfaces` is updated properly and also runtime network state is correctly updated:
````
$ netstat -rn
Kernel IP routing table
Destination     Gateway         Genmask         Flags   MSS Window  irtt Iface
0.0.0.0         10.0.2.2        0.0.0.0         UG        0 0          0 eth0
10.0.2.0        0.0.0.0         255.255.255.0   U         0 0          0 eth0
10.10.20.0      0.0.0.0         255.255.255.0   U         0 0          0 eth1
10.10.134.0     10.10.20.128    255.255.255.128 UG        0 0          0 eth1
10.10.138.0     10.10.20.138    255.255.255.192 UG        0 0          0 eth1
169.254.0.0     0.0.0.0         255.255.0.0     U         0 0          0 eth0
```

kris-lab:
@njam maybe you could comment here as well?

njam:
Philipp you know the code better, I'll leave it to you.

ppp0:
I wonder what happens if you first install a `/etc/network/interfaces` file as we have it when a machine is freshly deployed. IMO the Augeas parser can't always correctly change the values?
In such a file we have e.g. this (valid) construct:
```
[...]
up route add -net 10.0.0.0/8 gateway 10.55.40.129
down route del -net 10.0.0.0/8
up route add -net 161.26.0.0/16 gateway 10.55.40.129
down route del -net 161.26.0.0/16
```

But what we do with `route_opts` is `up route` with a one-liner... May be we should introduce `route_opts` as a list of routes each of which is resolved into a `up` and `down` line?

kris-lab:
Good point! If you have multi line configuration and you wanna replace with single line (we do now) then of course augeas will not match it. It uses "lenses" for single line matching, but not validate the configuration as the whole setup.

It actually raise a new question if config file should be purged while we apply augeas?

I like your suggestion to introduce array or hash for keeping `route_opts`.

kris-lab:
As discussed with @ppp0 the problem seems to be not reproducible.

kris-lab:
paired with @ppp0

PR #1552, opened 2016-10-24
Catch TrasportException in Satis Daemon

ppp0:

PR #1555, opened 2016-10-31
Catch TrasportException in Satis Daemon, take 2 (or 3?)

ppp0:
Follow-up of #1552


ppp0:
self-merging cc @kris-lab

PR #1556, opened 2016-11-01
Use Squid instead of polipo

ppp0:
To see if we get less network errors when running tests


ppp0:
@kris-lab if this passes, then I would propose to test on full build in CI (by merging this). If all is good, we can remove polipo in a second step. wdyt?

modules/squid_deb_proxy/manifests/init.pp
```diff
@@ -0,0 +1,40 @@
...
+
+  include 'squid_deb_proxy::service'
+
+  file {
+    '/etc/squid-deb-proxy/':
+      ensure => directory,
+      owner  => '0',
+      group  => '0',
+      mode   => '0644';
+    '/etc/squid-deb-proxy//mirror-dstdomain.acl.d/':
```
kris-lab: typo `/`?

kris-lab:
Ok, I see your points. I have just read recommendation why not to use `polipo` or `squid` itself and I agree.

PR #1557, opened 2016-11-01
Replace httpredir as repo source

ppp0:
Using [httpredirector](https://github.com/rgeissert/http-redirector) as deb repo source as we do makes caching very hard if not impossible (see [caching issues](https://github.com/rgeissert/http-redirector/search?q=cache&type=Issues&utf8=%E2%9C%93) )

We want to populate the cache with predictable urls - otherwise its usefulness will be impaired


ppp0:
This will only apply if `facter` states that the virtualization being run on is "virtualbox" - so it should not affect production (unless we run virtualboxes in production)

@kris-lab please review

kris-lab:
As discussed it looks good.

kris-lab:
But lets fix this as well ` apt-key adv --keyserver keyserver.ubuntu.com --recv-keys EEA14886`

ppp0:
Yes, this was due to a restrictive acl (fixed in #1558 )

PR #1560, opened 2016-11-02
Share cache log directory on Vagrantfile

ppp0:


ppp0:
This is needed to inspect (and store) the inner workings (logs) of the VM's proxy cache.
@kris-lab please review

kris-lab:
Why we need to store these logs?

ppp0:
The logs are otherwise stored (and disposed of) inside the VM  at every spec run. Impossible to debug like this

kris-lab:
That's actually my question, is it needed only for current debugging or future wise?

kris-lab:
code looks good, but I am not sure about value of this only?

ppp0:
I noticed that a spec that *always* failed (on ci5), did not fail at several attempts after having enables the (shared) log. Also, I'd like to observer how well the cache is being hit or missed (first observations are not encouraging, ie high miss rate). Can't we remove it a later stage if we have overcome the caching problem?

kris-lab:
ok!

PR #1565, opened 2016-11-15
Revert "Make iptables::entry persistent"

ppp0:
Reverts cargomedia/puppet-packages#1508


kris-lab:
@ppp0 should we open a follow up for #1508 to keep a track of it?

ppp0:
Good point, just opened one

PR #1566, opened 2016-11-15
Fix squid-proxy not caching debs

ppp0:


ppp0:
I found out that by allowing everything to be cached and accessed (see mirrors and cache directives in squid-deb-proxy.conf) debs actually do get cached, which of course massively increases testing performance. Let's see if CI can exploit those changes, too

@kris-lab please review

ppp0:
Caching is now working but we have another problem: https://github.com/cargomedia/puppet-packages/issues/1567

ppp0:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1594

PR #1568, opened 2016-11-16
Update mongodb key-server url

kris-lab:
Part of https://github.com/cargomedia/vagrant-boxes/pull/84


kris-lab:
@ppp0 wdyt about this change? During `vagrant-boxes` build I have found this very problematic to use `TCP/11371`.

ppp0:
If there is a real improvement, as you allegedly have noticed, then I'm fine with that. Using port 11371 seems to be a controversial issue anyway, see https://tools.ietf.org/html/draft-shaw-openpgp-hkp-00#section-2

kris-lab:
retest this please

PR #885, opened 2016-11-21
Retry specs in PR mode?

ppp0:
Because of network failures etc single specs can fail.
How about we retry specs X times during pull-request mode?

### DIFF
--- a/build-ci.sh
+++ b/build-ci.sh
@@ -15,5 +15,5 @@ if [ -z "${ghprbTargetBranch}" ]; then
 else
     # Pull request build
     git fetch origin
-    bundle exec rake spec:changes_from_branch[origin/${ghprbTargetBranch}]
+    no_retries=true bundle exec rake spec:changes_from_branch[origin/${ghprbTargetBranch}]
 fi
--- a/build-ci.sh
+++ b/build-ci.sh
@@ -11,9 +11,9 @@ bundle exec rake lint
 vagrant box update
 if [ -z "${ghprbTargetBranch}" ]; then
     # Full project build
-    bundle exec rake spec
+    retries=2 bundle exec rake spec
 else
     # Pull request build
     git fetch origin
-    no_retries=true bundle exec rake spec:changes_from_branch[origin/${ghprbTargetBranch}]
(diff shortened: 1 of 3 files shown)


ppp0:
@kris-lab please review

kris-lab:
Maybe I am missing something, but what about combining `retries` and `no_retries`. It looks like `no_retries=false` is basically `retries=0`? Then the default would be `retries=0` and only for CI it would be positive?

ppp0:
Fully agree, @kris-lab please re-run reviewing task

build-ci.sh
```diff
@@ -11,7 +11,7 @@ bundle exec rake lint
 vagrant box update
 if [ -z "${ghprbTargetBranch}" ]; then
     # Full project build
-    bundle exec rake spec
+    retries=2 bundle exec rake spec
```
kris-lab: maybe `retries` should go to the end of line, I think it will be cleaner, wdyt?

spec/lib/spec_runner.rb
```diff
@@ -189,7 +190,12 @@ def run
         spec.puppet_module.supported_os_list.each do |os|
           next unless @filter_os_list.nil? or @filter_os_list.include? os
           emit(:output, "Running #{spec.name} for #{os}\n".bold)
```
kris-lab: This line seems to be repeated twice: on line 192 and line 197?

kris-lab:
I have added small comments. The rest looks and works fine for me.

build-ci.sh
ppp0: I've been having issues at times when programs do not recognise environment variables as such but rather think they're arguments. It depends on how the program reads it's ARGV list. That's the reason why I alwasy like to prepend ENV variables. OTOH, I believe `bundle` and `rake` will handle this properly. Wdty?

kris-lab: ok!

spec/lib/spec_runner.rb
kris-lab: ok

kris-lab:
ok lgtm

PR #1577, opened 2016-11-23
[bug] vagrant provisioning on snapshot restore

ppp0:


ppp0:
A snapshot is taken from a provisioned box so that re-provisioning only can harm

@kris-lab please review

kris-lab:
lvgtm

PR #1583, opened 2016-12-02
Move ssl nginx config to main config, enable http2

njam:
Addressing https://github.com/cargomedia/puppet-cargomedia/issues/860

HTTP/2 will not work in Chrome because we have OpenSSL 1.0.1 but need 1.0.2, see
https://www.nginx.com/blog/supporting-http2-google-chrome-users/

### DIFF
--- a/modules/nginx/metadata.json
+++ b/modules/nginx/metadata.json
@@ -9,10 +9,6 @@
     {
       "operatingsystem": "Debian",
       "operatingsystemrelease": ["8"]
-    },
-    {
-      "operatingsystem": "Ubuntu",
-      "operatingsystemrelease": [ "15.04" ]
     }
   ],
   "dependencies": [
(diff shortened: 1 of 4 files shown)


njam:
@ppp0 @kris-lab please review

kris-lab:
- should `ssl_prefer_server_ciphers` be disbaled and allow the client to enforce?
- specs need an adjustments

otherwise it looks good

njam:
I've removed `ssl_prefer_server_ciphers`, so it will use the default - `off`.
The reason is that otherwise we need to maintain a list of ciphers (or use nginx' default list), which should be updated according to what browsers we want to support and what security constraints we have. Maintaining such a list is difficult.

ppp0:
>I've removed ssl_prefer_server_ciphers, so it will use the default - off.

While I understand the concerns, it's against most recommedations about security practices, see e.g. https://raymii.org/s/tutorials/Strong_SSL_Security_On_nginx.html#Extra_settings or https://goo.gl/N0q05s - Just as a note. With this setting off, it's *the browser* that will decide what cipher to be used

njam:
Yes I'm aware of that.
I researched what is a list of recommended ciphers to use, and it's a difficult question. It depends on what browsers you want to support in what versions, and what constraints you have on security. This requires quite some research and compromises to be made.
For our use case I think it's fine to let the client select a cipher from the list.

PR #1582, opened 2016-12-02
Add new mongodriver

alexispeter:
Github: https://github.com/mongodb/mongo-php-driver
Required version: >=v1.2.0


alexispeter:
@kris-lab could you review this and maybe help me fix/understand some remaining issues with the installation script?

1. There seem to be issues with permissions, as the install-script cannot create the files it needs. Running all command with sudo "fixes" this. This might also explain all the other failing modules in the ci-build.
2. Compiling the mongodb-driver requires openssl and pkg-config, however when the script gets to this point, it can't find openssl, which is there, because pkg-config is not there, although it's required in the manifest as part of the build-module.

kris-lab:
Yes, I think the problem is missing `pkg-config`. The solution seems to be in providing `--with-openssl-dir=/path/to/openssl`. Afaics it fails pretty early on the `./configure` step. 

We have two options:
- provide this path manually within `install.sh` by appending to the `./configure --with-openssl-dir=/path/to/openssl` or better
- require `build::pkg_config` manifest (simply add after `require 'build'`

For me it works locally, can you confirm if this work for you?

kris-lab:
@alexispeter there is still problem with spec, which will affect your local build. This is due https://github.com/cargomedia/puppet-packages/issues/1587

You can simply workaround this problem by commenting few lines in https://github.com/cargomedia/puppet-packages/blob/master/spec/hiera.json

e.g. by replacing the content with:
```
{
  "apt::update::options": [
    "Acquire::http::proxy=false"
  ],
  "helper::script::environment": [
//    "http_proxy=http://localhost:8123/",
//    "HTTP_PROXY=http://localhost:8123/",
//    "https_proxy=http://localhost:8123/",
//    "HTTPS_PROXY=http://localhost:8123/",
    "no_proxy=127.0.0.1,localhost",
    "NO_PROXY=127.0.0.1,localhost"
  ]
}
```

alexispeter:
Oh I see, I just assumed it would be automatically required with the build-module, but looking at the build-module's manifest it makes sense that it wouldn't.

kris-lab:
Could you please merge upstream. HTTP-proxy fix is already merged. I guess, this is the only problem why the build fails.

alexispeter:
please review and merge @kris-lab

kris-lab:
lgtm

What is the next step? Should we remove old `php5::extension::mongo` extension and update [cm::application](https://github.com/cargomedia/puppet-packages/blob/addf527fca6c26338ea258d9fd8899440bd4b3a1/modules/cm/manifests/application.pp#L19)

alexispeter:
Not quite yet. There are some issues left to deal with before we can switch.
The reason I already added this to puppet was so I could run jenkins on the sk-update-branch.

njam:
Could we have both extensions as dependencies of `cm::application` for some time? And once the migration is done we could remove the old one?

kris-lab:
I haven't tested that scenario. Have you @alexispeter ?

alexispeter:
Those are different extensions, so having them both active at the same time poses no problems for the application. Or in what way would you expect complications that would need to be tested?

kris-lab:
ok, sgtm

PR #1584, opened 2016-12-02
Install nginx from jessie-backports with OpenSSL 1.0.2

njam:
Addressing https://github.com/cargomedia/puppet-cargomedia/issues/860

@kris-lab please review

PR #1586, opened 2016-12-05
Install nginx from jessie-backports with OpenSSL 1.0.2

njam:
Addressing cargomedia/puppet-cargomedia#860
Tried before: https://github.com/cargomedia/puppet-packages/pull/1584

Problem: `class php5` installs `php5-dev` which requires `libssl-dev`, which we would also need to install from backports.

Notes to revert:
```
sudo rm -f /etc/apt/sources.list.d/backports.list
sudo rm -f /etc/apt/preferences.d/{libssl,nginx}*.pref
sudo apt-get update

sudo apt-get remove -y nginx-common nginx-full
sudo apt-get install  -y libssl1.0.0=1.0.1t-1+deb8u5
```


njam:
Here's a possible solution. Installing the *openssl* package from backports (which has version 1.0.2) on our `openssl` class and then require that. There are two downsides of this approach:
- Many other packages require openssl. Since we usually don't require the `openssl` class they will install the normal jessie *openssl* package in version 1.0.1. For those classes where we require openssl - like nginx - we will install 1.0.2 system-wide.
- We install backports on all machines which have nginx.

Alternative approaches would be:
- Compile nginx ourselves and link it statically against openssl 1.0.2. Downside: we need to manage and upgrade that compiled version (DEB?).
- Use the [nginx Docker container that runs on Alpine linux](https://hub.docker.com/_/nginx/). Downside: we need to install and manage Docker on production.
- Wait for the next Debian version

@cargomedia/devops thoughts?

kris-lab:
> We install backports on all machines which have nginx.

👎 I am bit scary about all dependencies. It should be backward compatible, but I see there are tons of packages which require ssl.

> We install backports on all machines which have nginx.

👎 it is safer but still not the perfect solution

> Compile nginx ourselves and link it statically against openssl 1.0.2. Downside: we need to manage and upgrade that compiled version (DEB?).

👍 👍 

> Use the nginx Docker container that runs on Alpine linux. Downside: we need to install and manage Docker on production.

👍 this is very time consuming now, but there will be a day when it makes the most of sense. 

Just as a note, maybe in reference to [special names](https://github.com/cargomedia/puppet-cargomedia/issues/1657) we should introduce load balancer (lb1) and not www (master role) but rather only `app1, app2, app3` where the traffic comes from `lb1`. This load balancer could possibly take case also of stream and any other global-ip traffic.

ppp0:
Probably, if we really want this particular flavour of nginx, the best (or least problematic?) would be to offer our own DEB

Is this the right repo to discuss so many internals?

kris-lab:
> Is this the right repo to discuss so many internals?

correct, we shouldn't in this repo!

njam:
agree

PR #1581, opened 2016-11-30
[experimental] - completely switch off proxy

ppp0:

PR #1587, opened 2016-12-05
NULL http-proxy for requests

ppp0:
I believe there is an issue with cache/proxy introduced in https://github.com/cargomedia/puppet-packages/pull/1581 by disabling proxy fully.

From my experience it interferes with spec setting https://github.com/cargomedia/puppet-packages/blob/master/spec/hiera.json

We have `env` defined to use polipo, @ppp0 can you investigate?


ppp0:
@kris-lab let's remove this hiera stuff to really get specs running *without* cache.

ppp0:
self-merging, cc @kris-lab

PR #1575, opened 2016-12-07
Make iptables::entry persistent, take 2

ppp0:
Follow-up of #1508


ppp0:
I found out that `ufw`'s `before.rules` syntax  basically is `iptables-save` or `restore` (see http://www.iptables.info/en/iptables-save-restore-rules.html)

The mistake we were making is *resetting* the chain. We only need to *add* to the `ufw-before-input` for `ufw` and `iptables` to properly handle our rules.

@kris-lab please review

ppp0:
How to proceed after merging:

- Stop puppet `systemctl stop puppet` everywhere
- On bulldog2 later on  stream4:
  - `systemctl stop ufw` (will invoke `ufw_stop` in `/lib/ufw/ufw-init-functions` which will clear all ufw managed chains and rules
  - Exec this to completely flush `iptables`:
  
  ```
  iptables -P INPUT ACCEPT
  iptables -P FORWARD ACCEPT
  iptables -P OUTPUT ACCEPT
  iptables -t nat -F
  iptables -t mangle -F
  iptables -F
  iptables -X
  ```
  - `puppet agent -t`

* if everything is good, do www8 manually

* restart puppet everywhere

ppp0:
paired w/ @kris-lab

PR #1588, opened 2016-12-08
[bug] Php5::Extension::Mongodb installer runs at every puppet run

ppp0:
Regression introduced with https://github.com/cargomedia/puppet-packages/pull/1582


ppp0:
Double-checked to be ok, merging, cc @kris-lab @alexispeter

PR #1589, opened 2016-12-08
Re-add polipo and restart it at every snapshot restore

ppp0:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1594

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,5 +1,17 @@
 Vagrant.configure('2') do |config|
-
+  if Vagrant.has_plugin?('vagrant-proxyconf')
+    config.vm.synced_folder '.proxy-cache', '/tmp/proxy-cache', :owner => 'proxy', :create => true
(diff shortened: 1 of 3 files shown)


ppp0:
I'm now merging this and starting a full build on ci
cc @kris-lab

ppp0:
vagrant destroy everywhere needed to enable the cache

njam:
You're restarting polipo in the puppet provisioner. I thought provisioners are not running on "snapshot go"?

ppp0:
true, we switched that provisioning off :( - I was originally aiming at the shell provisioner, actually, puppet should not do anything if the state's not modified. 

The shell script, though, has been changed. Now we check if puppet is installed (https://github.com/cargomedia/puppet-packages/blob/master/scripts/puppet-install.sh#L15) and if it is, nothing happens. We were observing issues with a run condition for dpkg which should not be the case anymore..

Let's remove `--no-provision` from `snapshot restore``

PR #1591, opened 2016-12-09
add openssl dependency

alexispeter:


alexispeter:
please review @njam

njam:
thanks!

cc @fauvel

PR #1603, opened 2016-12-15
Make iptables persistent, take 3

kris-lab:
Cont of https://github.com/cargomedia/puppet-packages/pull/1598

TODO:
- [x] make sure UFW works without `ipv6` module in kernel

PR #1608, opened 2016-12-19
Configure cm::vhost and cm::services::* to support UFW configs

kris-lab:
Part of #1603

### DIFF
--- a/modules/cm/manifests/services/stream.pp
+++ b/modules/cm/manifests/services/stream.pp
@@ -43,7 +43,7 @@ class cm::services::stream(
     statusPort  => $status_port,
   }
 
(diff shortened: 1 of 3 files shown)


kris-lab:
@njam please have a look. I think this could be improved, however I was under time pressure!

njam:
The reason is that we have UFW on all machines now?
- services/stream.pp: is *socket_ports* needed? Isn't this just internal for nginx?
- services/stream.pp: not sure if *status_port* should be open. In our case it should be open only on the "private" network.

Maybe open a new ticket to discuss?

kris-lab:
See continuation in https://github.com/cargomedia/puppet-packages/issues/1610

PR #1610, opened 2016-12-21
[UFW] Re-configure socket-redis rules

kris-lab:
Part of https://github.com/cargomedia/puppet-packages/pull/1608

> copying @njam 
The reason is that we have UFW on all machines now?
services/stream.pp: is socket_ports needed? Isn't this just internal for nginx?
services/stream.pp: not sure if status_port should be open. In our case it should be open only on the "private" network.
Maybe open a new ticket to discuss?

### DIFF
--- a/modules/cm/manifests/services/stream.pp
+++ b/modules/cm/manifests/services/stream.pp
@@ -44,6 +44,6 @@ class cm::services::stream(
   }
 
   @ufw::application { 'cm-services-stream':
-    app_ports => inline_template("${port},<%= @socket_ports.join(',')%>,${status_port}/tcp"),
+    app_ports => inline_template("${port},${status_port}/tcp"),
   }
 }


kris-lab:
@njam please have a look on comment above and review

kris-lab:
self-merging

njam:
> disagree, we use URI (with public IP) in application config for status port 

Hmmmmm true. But this endpoint should not be available on the public internet. Actually it isn't, but why? Because the node is behind the hardware firewall?

kris-lab:
Yes, the node is behind HW firewall. We use Public/Global-IPs owned by Softlayer, so the routing is done by them internally. This is of course the reason why `appX` servers can access `stream4` using Global-IP atm.

wdyt?

njam:
It's quite complicated and opaque. And it will only work as long as the app servers are behind the same firewall.

How about we make it *really* public, but introduce a "secret" that needs to be provided to access the status page?
If you agree, could you open a meta ticket on puppet-cargomedia? It should be easy to implement the functionality in socket-redis.

kris-lab:
Agree, it works fine only in this specific network setup.

Cont. in https://github.com/cargomedia/puppet-cargomedia/issues/1702

PR #1618, opened 2017-01-09
Add UFW rules for nginx::vhost resource

kris-lab:


kris-lab:
@njam please review

modules/nginx/manifests/resource/vhost.pp
```diff
@@ -77,6 +82,11 @@
       content => template("${module_name}/vhost/vhost_footer.erb"),
       notify  => Class['nginx::service'],
     }
+
+    @ufw::rule { "allow ssl for ${name}":
```
njam: let's call it "https" instead of "ssl"?

PR #1619, opened 2017-01-10
Add cdn-origin to cm:reverse_proxy

kris-lab:


kris-lab:
@njam please have a look

njam:
Could $cdn_origin be added to $hostnames? Seems the config of the vhost is exactly the same.

PR #1622, opened 2017-01-11
Use virtual resource for systemd::target

njam:


tomaszdurka:
Should we realize other systemd resources? Unit, service, other?

tomaszdurka:
Otherwise lgtm

njam:
added them!

PR #1629, opened 2017-01-16
tune2fs to disable fsck at boot

ppp0:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1669

### DIFF
--- a/modules/fs/Puppetfile
+++ b/modules/fs/Puppetfile
@@ -1,0 +1,7 @@
+mod 'cargomedia/apt',
+    :git => 'git@github.com:cargomedia/puppet-packages.git',
+    :path => 'modules/apt'
+
+mod 'cargomedia/helper',
+    :git => 'git@github.com:cargomedia/puppet-packages.git',
+    :path => 'modules/helper'
--- a/modules/fs/Puppetfile
+++ b/modules/fs/Puppetfile
@@ -2,6 +2,6 @@ mod 'cargomedia/apt',
     :git => 'git@github.com:cargomedia/puppet-packages.git',
     :path => 'modules/apt'
 
-mod 'cargomedia/helper',
+mod 'cargomedia/systemd',
     :git => 'git@github.com:cargomedia/puppet-packages.git',
-    :path => 'modules/helper'
+    :path => 'modules/systemd'
--- a/modules/fs/manifests/config/nofsck.pp
+++ b/modules/fs/manifests/config/nofsck.pp
@@ -2,9 +2,22 @@ class fs::config::nofsck {
 
   ensure_packages(['util-linux'], { provider => 'apt' })
 
-  helper::script { 'Disable fsck on all partitions':
+  file { '/usr/local/bin/disable-fsck-at-mount.sh':
+    ensure  => file,
     content => template("${module_name}/disable-fsck-at-mount.sh"),
-    unless  => 'ls -1 /var/local/nofsck',
+    owner   => '0',
+    group   => '0',
+    mode    => '0750',
     require => Package['util-linux'],
   }
+
+  $unit_name = 'nofsck.service'
+
+  systemd::unit { $unit_name:
+    service_name => $unit_name,
+    critical => false,
+    content => template("${module_name}/nofsck.service"),
+    require => File['/usr/local/bin/disable-fsck-at-mount.sh'],
+  }
+
 }
--- a/modules/fs/manifests/config/nofsck.pp
+++ b/modules/fs/manifests/config/nofsck.pp
@@ -13,6 +13,10 @@ class fs::config::nofsck {
 
   $unit_name = 'nofsck.service'
 
+  service { $unit_name:
+    enable => true,
+  }
+
   systemd::unit { $unit_name:
     service_name => $unit_name,
     critical => false,
--- a/modules/fs/manifests/config/nofsck.pp
+++ b/modules/fs/manifests/config/nofsck.pp
@@ -1,0 +1,10 @@
+class fs::config::nofsck {
+
+  ensure_packages(['util-linux'], { provider => 'apt' })
+
+  helper::script { 'Disable fsck on all partitions':
+    content => template("${module_name}/disable-fsck-at-mount.sh"),
+    unless  => 'ls -1 /var/local/nofsck',
+    require => Package['util-linux'],
+  }
+}
--- a/modules/fs/metadata.json
+++ b/modules/fs/metadata.json
@@ -1,0 +1,22 @@
+{
+  "name": "cargomedia-fs",
+  "version": "0.0.1",
+  "author": "Cargo Media",
+  "license": "MIT",
+  "summary": "Filesystem defaults",
+  "source": "https://github.com/cargomedia/puppet-packages",
+  "operatingsystem_support": [
+    {
+      "operatingsystem": "Debian",
+      "operatingsystemrelease": ["8"]
+    },
+    {
+      "operatingsystem": "Ubuntu",
+      "operatingsystemrelease": [
+        "15.04"
+      ]
+    }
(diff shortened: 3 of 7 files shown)


ppp0:
@kris-lab please review

Not sure about a couple of things. First and foremost on how to build the `unless` clause. I now opted for the creation of a status file in `/var/local` - This is more efficient than having to invoke `tune2fs` for every partition and evaluate its output

ppp0:
Somehow I tend to introduce a `systemd` unit that fires just before the machine is being shutdown/rebooted (there is a differentiation between shutdown and restart!).

The rationale is to ensure that the partitions are surely being `tune2fs`d before rebooting. I can think of edge cases where having puppet run the script and check for `/var/local/nofsck` would fail. One example is some kind of reconfiguration/addition/repartitioning of the drives where `/var/local/nofsck`  wouldn't be deleted.

Draft idea:
The unit in question would possibly be named `/etc/systemd/system/nofsck.service`, be of `type=oneshot` and be invoked `before=umount.target` (seehttps://www.freedesktop.org/software/systemd/man/systemd.special.html#umount.target`)

kris-lab:
Current implementation (without systemd) looks ok, but the downside is the time gap between reboot and puppet run to apply the required state?

`Systemd` service sounds good, but not sure if this is reliable solution?

ppp0:
Spec fails because of `modules/fs/spec/config-nofsck/manifest.pp - ERROR: Syntax error (try running `puppet parser validate <file>`) on line 5` not recognizing the new Puppet syntax allowing Heredocs

ppp0:
I tested with resetting the tune2fs params, reboot, read params.
Result:

```
root@ubuntu-1504-amd64-plain:/home/vagrant# systemctl status nofsck
● nofsck.service - Disable fsck at mount
   Loaded: loaded (/etc/systemd/system/nofsck.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Fri 2016-12-09 13:19:54 UTC; 1 months 8 days ago
  Process: 182 ExecStop=/usr/local/bin/disable-fsck-at-mount.sh (code=exited, status=0/SUCCESS)
  Process: 175 ExecStart=/bin/true (code=exited, status=0/SUCCESS)
 Main PID: 175 (code=exited, status=0/SUCCESS)

Dec 09 13:19:54 ubuntu-1504-amd64-plain disable-fsck-at-mount.sh[182]: tune2fs 1.42.12 (29-Aug-2014)
Dec 09 13:19:54 ubuntu-1504-amd64-plain disable-fsck-at-mount.sh[182]: Setting maximal mount count to -1
Dec 09 13:19:54 ubuntu-1504-amd64-plain disable-fsck-at-mount.sh[182]: Setting interval between checks to 0 seconds
Warning: Journal has been rotated since unit was started. Log output is incomplete or unavailable.
root@ubuntu-1504-amd64-plain:/home/vagrant# reboot
Shared connection to 127.0.0.1 closed.
Connection to 127.0.0.1 closed by remote host.
cargomedia@dog:~/Projects/puppet-packages (issue-1629)$ ssh1504
Welcome to Ubuntu 15.04 (GNU/Linux 3.19.0-58-generic x86_64)

 * Documentation:  https://help.ubuntu.com/
Your Ubuntu release is not supported anymore.
For upgrade information, please visit:
http://www.ubuntu.com/releaseendoflife

New release '16.04.1 LTS' available.
Run 'do-release-upgrade' to upgrade to it.

Last login: Tue Jan 17 14:23:22 2017 from 10.0.2.2
vagrant@ubuntu-1504-amd64-plain:~$ sudo su
root@ubuntu-1504-amd64-plain:/home/vagrant# systemctl status nofsck
● nofsck.service - Disable fsck at mount
   Loaded: loaded (/etc/systemd/system/nofsck.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Fri 2016-12-09 13:23:00 UTC; 1 months 8 days ago
  Process: 193 ExecStop=/usr/local/bin/disable-fsck-at-mount.sh (code=exited, status=0/SUCCESS)
  Process: 167 ExecStart=/bin/true (code=exited, status=0/SUCCESS)
 Main PID: 167 (code=exited, status=0/SUCCESS)

Dec 09 13:23:00 ubuntu-1504-amd64-plain disable-fsck-at-mount.sh[193]: tune2fs 1.42.12 (29-Aug-2014)
Dec 09 13:23:00 ubuntu-1504-amd64-plain disable-fsck-at-mount.sh[193]: Setting maximal mount count to -1
Dec 09 13:23:00 ubuntu-1504-amd64-plain disable-fsck-at-mount.sh[193]: Setting interval between checks to 0 seconds
Warning: Journal has been rotated since unit was started. Log output is incomplete or unavailable.
```

The script has been run. Question is: was it run before or after rebooting?

ppp0:
Alright, @kris-lab please re-review.
I changed the service definition and tested it. It now runs at install time and before rebooting (thus modifying the settings as expected). There is one thing to do: enable the unit.. Since there is no (puppet) service defined, the unit is not being enabled as in https://github.com/cargomedia/puppet-packages/blob/master/modules/systemd/manifests/unit.pp#L26 - Shall I define a service resource even though we have a oneshot script here?

kris-lab:
As discussed, I would enable this service like we do e.g. with `systemd::critical_units`.

njam:
Remember that we also might have to hard reset a server, in what case I guess that systemd unit will not execute.
The unit's description is "Disable fsck at mount". If we can execute it on mount that would be perfect wouldn't it? Is it like that?

ppp0:
As discussed, there is no way we can ensure 100% that fsck will be disabled. To elucidate what is meant:
* The service unit **will be executed at (puppet) install time**, thus modifying the mounted partiions - this is in most cases the ones we want
* ""Disable fsck at mount" means Disable "fsck at mount". Sorry if there is room for misunderstanding. The decision if a `fsck` has to be made is based on information and configuration residing on the partition's config sectors and will be taken "at mount time"

There is edge cases as for instance if we modify or mount additional partitions and/or drives without rebooting in which case said partition/drive won't be reconfigured the way we want. 

BTW, I randomly checked two nodes on SoftLayer (app2 and search3) and `tune2fs` returns these defaults:
```
[..]
Maximum mount count:      -1
[..]
Check interval:           0 (<none>)
[..]
```

which is exactly what we want

I would say it is worth to finish up this topic as it is now

njam:
I see, makes sense!

ppp0:
CI is failing because of puppet-lint and heredocs :(

Manual test to be done anyway:

* Apply puppet
* Manually reset (`tune2fs -c 30 -i 3m /dev/sda1`)
* Reboot
* Check settings (`tune2fs -l /dev/sda1`)

All passed

@kris-lab please review

kris-lab:
lgtm, but what we do with "puppet-lint and heredocs"? Do you know which part of formatting causes this?

ppp0:
Lint does not like this construct:
```
$reset_script = @(EOS)
  for fs in ext2 ext3 ext4; do
    for i in $(findmnt -lno source -t ${fs}); do  <---- line 5 is this
      tune2fs -c 30 -i 3m ${i}
    done;
  done;
    | EOS
```

````
modules/fs/spec/config-nofsck/manifest.pp - ERROR: Syntax error (try running `puppet parser validate <file>`) on line 5
````

IIRC I had a similar problem when pushing this: https://github.com/cargomedia/puppet-packages/blob/master/modules/ufw/spec/before/manifest.pp#L3

Let's see if I can get rid of the "syntax error" (I don't see why `puppet-lint` would care about the syntax.. different underlying puppet gem?

ppp0:
I changed the reset script. No need to go through all possibly existing extN partitions when we later only check `tune2fs -l /dev/sda1`

@kris-lab please review

PR #1631, opened 2017-01-18
Allow create replica-set without enabled sharding

kris-lab:


kris-lab:
@njam @ppp0 fyi, this is quick workaround to support replica-sets without full sharding. Alternative approach would be copy this class as e.g. `mongodb::role::replica_node` wihtout `mongodb_shard` section.

njam:
As discussed let's introduce param $enable_sharding

kris-lab:
self merging

PR #1634, opened 2017-01-19
Upgrade foreman to 0.83

njam:
Addressing https://github.com/cargomedia/puppet-cargomedia/issues/1646

@cargomedia/infrastructure please review

### DIFF
--- a/modules/foreman/manifests/init.pp
+++ b/modules/foreman/manifests/init.pp
@@ -1,6 +1,8 @@
-class foreman {
+class foreman(
+  $version = '0.83.0',
+) {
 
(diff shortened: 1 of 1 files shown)


ppp0:
lgtm - but you want to give up "automatic upgrades"?

njam:
We don't have automatic upgrades now, for that we would need to use "latest". Currently we have a "random version" (what was available during the initial puppet run).
I would not use "latest" here because it's a crucial part of our deployment process, and the software is not maintained very actively.

ppp0:
sorry for the confusion, as said, lgtm

PR #1640, opened 2017-01-25
Detect OS using /etc/os-release in puppet installer

njam:


njam:
@kris-lab @ppp0 wdyt?

I'm using the newer `/etc/os-release` to detect the operating system (instead of LSB). Originally I thought we don't have to install "lsb-release" at all, but unfortunately it seems not possible (Ubuntu has a file `/etc/lsb-release`, but not Debian).
I think it's still better to detect the OS using `os-release`, which is widely supported, and then use lsb-release to detect the codename.

@ppp0 please also comment on the first commit

ppp0:
Regarding the [first commit](https://github.com/cargomedia/puppet-packages/pull/1640/commits/ed00ee125dcc34b11f8147e8936fc11314ceb1c1):
The rationale was/is to avoid a redundant `apt-get update` if the box is freshly provisioned. We manage this file in https://github.com/cargomedia/puppet-packages/blob/master/modules/apt/manifests/update.pp#L14

ppp0:
rest, lgtm

njam:
Yes, did you see my comment in the commit message about it?
> This file is part of our puppet module, but writing it here is not technically needed and it's probably better to not rely on it.
The only upside I can think of is that the initial installation of a host that DOESN'T have puppet yet will be a bit faster, but that happens rarely.

wdyt? are there cases when we execute this script regularly on a machine that doesn't have puppet installed?

ppp0:
I did not see it, sorry.  lgtm

njam:
@kris-lab could you double check it too?

kris-lab:
lgtm, works for me

PR #1638, opened 2017-01-24
Add MongoDB storage engine param - WiredTiger support

kris-lab:
Depends on https://github.com/cargomedia/puppet-packages/issues/1255
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1747
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1641

Issue #1255, opened 2016-04-18
[Feature] Upgrade to mongodb 3.2


njam:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1747

Current version 2.6 is approaching end of life, and MMS will stop supporting it at the end of October, 2016.
https://www.mongodb.com/blog/post/mongodb-2-6-end-of-life

Also version 3 has some very interesting performance improvements.

kris-lab:
Test plan:
- [ ] restore `rep1` locally
- [ ] upgrade manually
- [ ] verify with SK dev app

kris-lab:
Upgrade recommendation https://docs.mongodb.com/v3.0/tutorial/install-mongodb-on-debian/

kris-lab:
Upgrade from 3.0 to 3.2
- [ ] https://docs.mongodb.com/manual/release-notes/3.2-upgrade/

njam:
Blocked by https://github.com/cargomedia/cm/pull/2331

kris-lab:
Follow up https://github.com/cargomedia/puppet-packages/pull/1639

PR #1641, opened 2017-01-25
Add support for multiple mongo versions

kris-lab:
Cont #1255
See also #1638


njam:
lgtm

not sure about all the dependencies, but specs will tell

PR #1642, opened 2017-01-26
BO-38 [nginx] Make access_log more configurable

ppp0:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1758 and https://github.com/cargomedia/puppet-cargomedia/issues/1755

### DIFF
--- a/modules/nginx/templates/vhost/vhost_header.erb
+++ b/modules/nginx/templates/vhost/vhost_header.erb
@@ -3,8 +3,8 @@ server {
 <% unless @server_name.empty? %>
   server_name           <%= @server_name.join(" ") %>;
 <% end %>
-<% if scope.lookupvar('nginx::params::nx_access_log') == 'on' %>
-  access_log            <%= scope.lookupvar('nginx::params::nx_logdir')%>/<%= @name %>.access.log;
+<% if scope.lookupvar('nginx::params::nx_access_log') != 'off' %>
+  access_log            <%= scope.lookupvar('nginx::params::nx_access_log') %>;
 <% end %>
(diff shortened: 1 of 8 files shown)


ppp0:
@kris-lab please review

modules/nginx/templates/vhost/vhost_header.erb
```diff
@@ -3,8 +3,8 @@ server {
 <% unless @server_name.empty? %>
   server_name           <%= @server_name.join(" ") %>;
 <% end %>
-<% if scope.lookupvar('nginx::params::nx_access_log') == 'on' %>
-  access_log            <%= scope.lookupvar('nginx::params::nx_logdir')%>/<%= @name %>.access.log;
+<% if scope.lookupvar('nginx::params::nx_access_log') != 'off' %>
+  access_log            <%= scope.lookupvar('nginx::params::nx_access_log') %>;
 <% end %>
```
kris-lab: maybe update with `<% end -%>`  so the config will not change and service will not restart? Btw, should we set `nginx::params::nx_log_format` with `off` instead of `undef`?

kris-lab: Sorry, please ignore! I mixed variables.

kris-lab:
@ppp0 code looks good. If you think we still need this, I suggest to merge this. Otherwise let's close.

ppp0:
We're not using it *for the moment* - but we might in the future (thinking of https://github.com/cargomedia/puppet-cargomedia/issues/1411) - OK to merge, nothing should be affected

Issue #1649, opened 2017-01-31
nginx::resource::{location,upstream,vhost}


ppp0:
There's a common argument `$ensure` which could be refactored away in all defines.

We don't really use `$ensure => absent` to remove resources, afaicr and it successfully contributes to unreadability without real benefits

e.g. 
```pp
define nginx::resource::upstream (
  $ensure = present,
 [..]
) {
 [..]
  $fileIfPresent = $ensure ? { present => file, default => $ensure }
  file { "/etc/nginx/conf.d/${name}-upstream.conf":
    ensure   => $fileIfPresent,
    [..]
  }
```

ppp0:
closed as solved

PR #1650, opened 2017-01-31
remove passing $ensure argument from all nginx::resources

ppp0:


ppp0:
resolves https://github.com/cargomedia/puppet-packages/issues/1649

ppp0:
@kris-lab please review

kris-lab:
Agree in general because we purge `conf.d`. We use atm:
```
    File["${nginx::params::nx_conf_dir}/conf.d"] {
      ignore  => 'vhost_autogen.conf',
      purge   => true,
      recurse => true,
    }
```

However we store partial file in `nginx.d` for concatenation in next step. Is it still properly working without setting `ensure => absent` for common resources (vhost, location, upstream)?

ppp0:
* Where do we use `ensure => absent` to remove resources? 
* About purging. We could even get rid of this  `$nx_confd_purge` mechanism, as the default is true (makes sense to me) and we don't have cases where we would not purge the `conf.d` dir?
* Concatenation happens in a temp dir `${nginx::config::nx_temp_dir}/nginx.d` which is being populated in `nginx::service` 
```
  exec { 'rebuild-nginx-vhosts':
    command     => "/bin/cat ${nginx::params::nx_temp_dir}/nginx.d/* > ${nginx::params::nx_conf_dir}/conf.d/vhost_autogen.conf",
    refreshonly => true,
    unless      => "/usr/bin/test ! -f ${nginx::params::nx_temp_dir}/nginx.d/*",
    subscribe   => File["${nginx::params::nx_temp_dir}/nginx.d"],
  }
```

kris-lab:
I haven't tested in person this new logic without param `ensure => [present/absent/file]` and with removed e.g.:
```
## Shared Variables
   $ensure_real = $ensure ? {
     'absent' => absent,
     default  => file,
   }
```
I was thinking that it could change the behaviour where incorrect temporary files from `nginx.d` could be concatenated into final `conf.d`.

If you tested all this cases then it looks good to me.

ppp0:
retest this please

PR #1667, opened 2017-02-15
Added required AWS version

fauvel:
Part of https://github.com/cargomedia/cm/pull/2576

### DIFF
--- a/modules/s3export_backup/manifests/init.pp
+++ b/modules/s3export_backup/manifests/init.pp
@@ -1,9 +1,10 @@
 class s3export_backup (
   $version = '0.3.0',
-  $aws_key,
-  $aws_secret,
+  $aws_version,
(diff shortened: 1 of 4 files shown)


njam:
I don't think this is needed. This config is for this application:
https://github.com/cargomedia/s3export_backup

It uses CM "~1.94.0". Until CM gets updated there, the config doesn't need to be changed?
Or is it backwards-compatible and you just want to do it already?

fauvel:
Ok, I didn't know about this app using an old version of cm. I'm not sure the change is backwards compatible, because this version of the AWS API didn't exist at the time the SDK v2 was released. I would close this issue then, agree?

njam:
Did you test it? As far as I can see we only add a config parameter, so I guess the parameter matching might just ignore it?

PR #1628, opened 2017-02-16
cm::vhost: simplify passing upstream options

ppp0:
Via https://github.com/cargomedia/docker-images/pull/6

Currently this code is *very* difficult to read:
```puppet
define cm::vhost(
[…]
  $upstream_options = { },
) {
[…]

  $upstream_options_defaults = {
    name => 'fastcgi-backend',
    members => ['localhost:9000']
  }

  $upstream_opts = merge($upstream_options_defaults, $upstream_options)
  $upstream_name_real = $upstream_options[name] ? { default => $upstream_opts[name], undef => $upstream_options_defaults[name] }

  if ($upstream_options[name] == undef) {
    if !(defined(Cm::Upstream::Fastcgi[$upstream_opts[name]])) {
      cm::upstream::fastcgi { $upstream_opts[name]:
        members => $upstream_opts[members]
      }
    }
  }
```

It allows to define the `cm::upstream::fastcgi` outside and passing only its "name" (in which case "members" will be ignored). Or we can not pass any options, then a default upstream will be created. As an edge case one could only pass "members" but no "name", in that case an upstream would get created with the members that were passed in.

I would suggest to change it to *always create the upstream inside the vhost* based on the params passed in. Also we should use `ensure_resource()` instead of `if !(defined()) { … }` which is more robust (errors when params differ from existing resource).

The only usages of `$upstream_options` I can find are `sk::www` and `sk::proxy`.

@kris-lab @ppp0 wdyt? could you do the change if you agree?
cc @zazabe


ppp0:
solved in #1669

PR #1669, opened 2017-02-16
Simplify cm::vhost cm::reverse_proxy UX

ppp0:
via #1628

### DIFF
--- a/modules/cm/manifests/reverse_proxy.pp
+++ b/modules/cm/manifests/reverse_proxy.pp
@@ -5,27 +5,16 @@ define cm::reverse_proxy(
   $aliases = [],
   $redirects = undef,
   $cdn_origin = undef,
-  $upstream_options = { },
+  $upstream_members = ['localhost:443'],
+  $upstream_protocol = 'https',
 ) {
 
   include 'nginx'
 
-  $upstream_options_defaults = {
-    name        => 'reverse-proxy-backend',
-    members     => ['localhost:443'],
-    ssl         => true,
-    header_host => '$host',
-  }
+  $backend_name = "${name}-proxy-backend"
 
-  $upstream_opts = merge($upstream_options_defaults, $upstream_options)
-  $upstream_name_real = $upstream_options[name] ? { default => $upstream_opts[name], undef => $upstream_options_defaults[name] }
-
-  if ($upstream_options[name] == undef) {
-    if !(defined(Cm::Upstream::Proxy[$upstream_opts[name]])) {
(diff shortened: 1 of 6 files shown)


ppp0:
@kris-lab please review

modules/cm/manifests/reverse_proxy.pp
```diff
@@ -65,9 +52,9 @@
       'real_ip_recursive on;',
       'real_ip_header X-Real-IP;',
       'set_real_ip_from 0.0.0.0/0;',
-      "proxy_set_header Host '${upstream_opts[header_host]}';",
+      'proxy_set_header Host $host;',
       'proxy_set_header X-Real-IP $remote_addr;',
-      "proxy_pass ${proto}://${upstream_name_real};",
+      "proxy_pass ${upstream_protocol}://${backend_name};",
```
kris-lab: I am wondering how it works if you specify proxy protocol as `https`, but upstream members are mix of `80/http` and `443/https`, do you know?

modules/cm/manifests/vhost.pp
```diff
@@ -1,38 +1,30 @@
...
-  $aliases = [],
-  $redirects = undef,
-  $cdn_origin = undef,
-  $debug = false,
-  $upstream_options = { },
+  $ssl_port          = 443,
+  $aliases           = [],
+  $redirects         = undef,
+  $cdn_origin        = undef,
+  $path,
```
kris-lab: Shouldn't we move all required params on the top of list?

kris-lab:
I like it and just wondering if we need to specify `upstream_protocol` and how it works with mixed protocols?

modules/cm/manifests/reverse_proxy.pp
ppp0: I don't think we have a use case in `cm::` for this at the moment. Is it possible at all? That would either require several `proxy_pass` in the same `location` context. Even if this is possible (may be by defining a mix of http and https backends in `upstream` it still looks to me like a hypothetical use case. No?

modules/cm/manifests/vhost.pp
ppp0: makes sense!

ppp0:
@kris-lab  please have a look again

modules/cm/manifests/reverse_proxy.pp
kris-lab: Agree, we do not implement such case for the moment.

kris-lab:
I think it good. Let's wait for build and apply carefully.

PR #1670, opened 2017-02-17
Re-enable systemd stdout to journal

kris-lab:


kris-lab:
paired with @njam

njam:
It will restart ALL services!
Let's do it manually for mysql.

PR #1681, opened 2017-02-28
BO-59 Ignore owner/group, set world x on cache dir, remove old/redundant chown

ppp0:


ppp0:
I found an old code fragment (see https://github.com/cargomedia/puppet-packages/pull/1278#issuecomment-281123958) which I'm removing now as redundant.
Let's not care about who is owning the cache directory, just about it's accessibility from bipbip

@kris-lab please review

kris-lab:
@ppp0 this logic for `cache` management feels very strange. I don't remember reason for that, but let's hope it will not break something!

I am ok to merge

PR #1682, opened 2017-03-02
BO-38 [bug] Remove ensure argument

ppp0:
Follow-up of https://github.com/cargomedia/puppet-packages/pull/1642


ppp0:
self-merging due to triviality.
Noticed that due to failing nightly build

@kris-lab cc

PR #1683, opened 2017-03-02
BO-107 Refactor cm::reverse_proxy to allow for mixed http(s) proxy chains

ppp0:


modules/cm/manifests/reverse_proxy.pp
```diff
@@ -24,27 +28,42 @@
...
+    nginx::resource::vhost { "${name}-https-redirect":
+      listen_port         => $listen_port,
+      server_name         => $hostnames,
+      location_cfg_append => [
+        'return 301 https://$host$request_uri;',
+      ],
+    }
+  }
+
+  # This construct is needed due to vhost considering itself SSL-only if
```
ppp0: `nginx::resource::vhost` needs `listen_port` and `ssl_port` to be equal to enforce SSL only (which in `cm::reverse_proxy` we handle in a separate variable to simplify things). This is the reason for that construct

modules/cm/manifests/reverse_proxy.pp
```diff
@@ -24,27 +28,42 @@
...
+    $protocol = $ssl_only_final ? {
+      true    => 'https',
+      default => '\$scheme',
+    }
+
+    nginx::resource::vhost { "${name}-redirect":
+      listen_port         => $listen_port,
       server_name         => $redirects,
       location_cfg_append => [
-        "return 301 https://${name}\$request_uri;",
```
ppp0: This one is about the redirects to the original hostname (e.g. `example.co.uk` redirects to `example.com`) - for that, if the host in NOT ssl-only, we need to adapt the redirection's protocol

modules/cm/spec/reverse-proxy/two-proxies/manifest.pp
```diff
@@ -1,8 +1,5 @@
 node default {
 
-  $upstream1_name = 'foobar'
```
ppp0: This is just enneeded - discovered by chance and related insofar as it is about reverse-porxy

ppp0:
@kris-lab please review

modules/cm/manifests/reverse_proxy.pp
```diff
@@ -1,17 +1,21 @@
...
+  $redirects         = undef,
+  $cdn_origin        = undef,
+  $upstream_members  = ['localhost:443'],
   $upstream_protocol = 'https',
 ) {
 
   include 'nginx'
 
   $backend_name = "${name}-proxy-backend"
+  $ssl_enabled = (is_integer($ssl_port) and $ssl_port > 0)
```
ppp0: Unfortunately, it is not possible to override a default (here: `ssl_port = 443`)  with `undef` !
So we need to pass either `ssl_port=0` or some other value such as `ssl_port = false` if we want to disable SSL

kris-lab: I am wondering if we need this validation at all. We don't do it in any other place, wdyt? 

It would simply a workflow so we assume, if `ssl_only` is `true`, then we do trust the value of `ssl_port`?

kris-lab:
In general, I am not sure if SSL should be 100% enforced atm. What if traffic comes from affiliates and other partners which points to plain http? 

I agree on enforcement of SSL for all interconnection and services communication, wdyt?

modules/cm/manifests/reverse_proxy.pp
ppp0: Sorry, I am confused..
You say:
> if ssl_only is true, then we do trust the value of ssl_port

This line, though, is about deciding *if* ssl is enabled *at all* - and I don't like to introcude `$ssl = [true|false]` such as in https://github.com/cargomedia/puppet-packages/blob/master/modules/nginx/manifests/resource/vhost.pp#L5

ppp0:
>if SSL should be 100% enforced atm

😕 ?

let's have a quick talk, wdyt?

ppp0:
As discussed, let's stop puppet on production servers, merge this, apply puppet with `--noop` and see what happens

PR #1686, opened 2017-03-08
BO-38 Add unit to run goreplay as a service

ppp0:


ppp0:
@kris-lab this is not yet finished, but with a bit of fine-tuning, we can use it to start traffic recording

kris-lab:
ok, sgtm

ppp0:
@kris-lab ready for review, please

kris-lab:
What is the reasoning to wrap this tool in the service? I was thinking to use it in supervised manner in the `screen` with dumping to the file.

It can be done as a service too, but we should maybe make it more configurable. Each service run should create new file with some date/time in the name?

ppp0:
>wrap this tool as a service

The reasoning is easy deployment and operation (a duration can be set). But I agree, it can be invoked manually, too. I left the args wide open so that one basically can inject ANY arguments. May be instead of `$input`, `$output`, `$duration` we could just have `$arguments` - The file naming scheme is defined via args

ppp0:
Also, this works as a testable PoC

PR #1687, opened 2017-03-09
BO-122 Manage runtime directory for socket files, use defaults everywhere

ppp0:


modules/mysql/manifests/server.pp
```diff
@@ -27,7 +27,7 @@
     before => Package['mysql-server'],
   }
 
-  file { '/etc/mysql':
+  file { ['/etc/mysql', '/var/run/mysqld']:
```
ppp0: We already manage `/etc/mysql`, so let's also manage `/var/run/mysqld`. Both are created by installing mysql through its `deb` - both have to exist for `mysql` to operate reliably

kris-lab:
retest this please

ppp0:
I fixed some things that made `mysqld` fail. Interestingly (bug/feature/old-school unixism?) `ExecStartPost=` is forked and exec just after `ExecStart=` is forked and exec'd thus running in parallel - big potential for collisions, especially in the situation we were having:

* `ExecStart=` fails
* `ExecStartPost=` runs for 30secs (waits for mysql to be up)
* Since it failed, it's being restarted
* etc.

I find it curious at least that `ExecStartPost` runs regardless of how `ExecStart` fared..

Anyway: the reason for failure of `mysqd` were a) an obsolete config directive b) no rights to write a UNIX socket to `/var/run/mysqld`

@kris-lab please review

kris-lab:
As discussed looks good. Let's apply carefully and verify.

Should we build new boxes?

ppp0:
I suggest to 
* stop puppet on prod mysql servers
* briefly stop app
* apply to slaves first
* apply to master
* start app
* start puppet

>Should we build new boxes?
I think this is a good idea as it hasn't been done since a while

kris-lab:
Plan sounds good. This change should not affect operations really, but restart must be fully orchestrated, agree.

PR #1689, opened 2017-03-16
BO-142 Upgrade upsource

ppp0:
screen shot 2017-03-16 at 09 50 04


ppp0:
self-merging due to triviality, cc @kris-lab

kris-lab:
You know than upgrade is fully manual process yes? You have to stop service, move data, install new version and then setup new installation by doing upgrade/recovery path.

PR #1690, opened 2017-03-20
BO-83 Add tmpfiles.d config to recreate /var/run/mysqld

ppp0:


ppp0:
See manpage for `tmpfiles.d(5)` on jessie (because of v215)

Like this, at boot time, the dir will be recreated with the correct perms. Actually, I think conf file this should go into the deb package!

modules/mysql/templates/tmpfiles.d/mysql.conf
```diff
@@ -0,0 +1,2 @@
+ # (Re-)Create the temp directory at boot time
+ d /run/mysql 0755 mysql mysql
```
kris-lab: What is exactly the source of the problem, `/var/run/mysqld` or `/run/mysql` ?

ppp0:
there was a typo. @kris-lab please review

kris-lab:
Shouldn't it be ` /var/run/mysqld` like configured https://github.com/cargomedia/puppet-packages/blob/master/modules/mysql/manifests/server.pp#L30 ?

ppp0:
@kris-lab of course it should! Thanks for reading carefully

ppp0:
I would suggest to proceed as usual when critical pieces are affected:
* stop puppet on all clustered dbs (ads, gearman, denk not affected)
* apply manually on one of the replicas to see if service gets a restart (should not!)
* restart puppet everywhere

I tested this locally and
* There is no "daemon" to be reloaded as in `daemon-reload`
* The config files in `tmpfiles.d` are read at boot time
See
```
# systemd-analyze blame
[..]
            45ms systemd-tmpfiles-clean.service
[..]
            39ms systemd-tmpfiles-setup-dev.service
[..]
            18ms systemd-tmpfiles-setup.service
[..]
```

PR #1697, opened 2017-04-13
BO-174 Remove fluentd logging, make all configs virtual, class instead of define for unique configs

ppp0:


modules/fluentd/manifests/init.pp
```diff
@@ -46,31 +46,17 @@
...
-    args          => '-c /etc/fluentd/fluent.conf -o /var/log/fluentd/fluentd.log --no-supervisor',
+    args          => '-c /etc/fluentd/fluent.conf --no-supervisor',
     user          => 'fluentd',
-    require       => [Ruby::Gem['fluentd'], File['/etc/fluentd/fluent.conf'], File['/var/log/fluentd'], User['fluentd']],
+    require       => [Ruby::Gem['fluentd'], File['/etc/fluentd/fluent.conf'], User['fluentd']],
   }
 
-  Fluentd::Config::Match <||>
-  Fluentd::Config::Source <||>
-  Fluentd::Config::Source_tail <||>
```
tomaszdurka: Why do we remove source/source_tail here?

ppp0: You are right, thanks. The rationale behind this was to simplify resource realisations. But let's better not touch source_tail as it's being used and would lead to reconfiguration

modules/fluentd/manifests/config/source_tail.pp
```diff
@@ -11,7 +11,7 @@
     type     => 'tail',
     config   => {
       path        => $path,
-      pos_file    => '/var/lib/fluentd/tail_pos',
+      pos_file    => "/var/lib/fluentd/${title}_pos",
```
tomaszdurka: shall we still keep `tail_` prefix? `tail_${title}_pos` not sure if it can conflict with anything...

ppp0: Agree

ppp0:
solved in #1699

PR #1699, opened 2017-04-27
BO-174 Fluentd: Add systemd-journald source class

ppp0:
Depends on #1697

### DIFF
--- a/modules/fluentd/spec/config/1-manifest.pp
+++ b/modules/fluentd/spec/config/1-manifest.pp
@@ -14,20 +14,23 @@ node default {
     },
   }
 
+  file { '/tmp/my-source2':
+    ensure   => file,
+    mode     => '0644',
+    group    => '0',
+    owner    => '0',
+  }
+
   fluentd::config::source_tail{ 'my-source-2':
     path        => '/tmp/my-source-2',
     fluentd_tag => 'source2',
+    require => File['/tmp/my-source2'],
   }
 
   ## Filter
 
-  fluentd::config::filter_add_hostname{ 'add-hostname':
-    pattern => '**',
-  }
-
-  fluentd::config::filter_streamline_level { 'streamline-level':
-    pattern => '**',
-  }
+  include 'fluentd::config::filter_add_hostname'
+  include 'fluentd::config::filter_streamline_levels'
 
   fluentd::config::filter{ 'my-filter-1':
     pattern  => 'filter1.**',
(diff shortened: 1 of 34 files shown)


njam:
retest this please

modules/fluentd/spec/config/1-manifest.pp
```diff
@@ -14,20 +14,23 @@
...
   ## Filter
 
-  fluentd::config::filter_add_hostname{ 'add-hostname':
-    pattern => '**',
-  }
-
-  fluentd::config::filter_streamline_level { 'streamline-level':
-    pattern => '**',
-  }
+  include 'fluentd::config::filter_add_hostname'
```
njam: Let's define the class here, so we can configure it (like it was done before with the define). There's no reason to "include" here?

modules/fluentd/spec/config/1-manifest.pp
```diff
@@ -14,20 +14,23 @@
...
+  include 'fluentd::config::filter_streamline_levels'
```
njam: Ditto

modules/fluentd/spec/source_journald/spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'spec_helper'
+
+describe 'fluentd:source-journald' do
+
+  describe service('fluentd') do
+    it { should be_running }
+  end
+
+  describe file('/var/lib/fluentd/journald_pos') do
```
njam: Why should this be a directory, we are not managing it are we?

modules/fluentd/spec/source_journald/spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it { should be_directory }
+  end
+
+  describe file('/tmp/dump') do
+    it { should be_directory }
+  end
+
+  describe command('journalctl -u fluentd |grep "fluentd worker is now running"') do
```
njam: How about adding a message to the journal specifically for the test (instead of relying on the message of fluentd which can change)?

modules/fluentd/spec/source_journald/spec.rb
ppp0: Because pos_file aka
```
<storage>
  @type local
  persistent true
  path /var/lib/fluentd/journald_pos
</storage>
```
 is somewhat misleading as it creates a directory with files in it. If it fails it does so silently and errors to the log, so this is a shortcut to check if it works

PR #1701, opened 2017-05-02
BO-199 Introduce a one-time unit to run swapoff -a

ppp0:


modules/fs/manifests/config/swapoff.pp
```diff
@@ -0,0 +1,20 @@
...
+    enable => true,
+  }
+
+  systemd::unit { $unit_name:
+    service_name => $unit_name,
+    critical => false,
+    content => template("${module_name}/swapoff.service"),
+  }
+
+  notify { $unit_name:
```
njam: What is this for?

ppp0: ? The Notify? To execute the "ExecStart" in the unit, otherwise it's only enabled, meaning it will be run at boot

ppp0: I did not want to say `System{ ensure => running, }` as it sounds weird?

ppp0:
I reworked the class. It's better to use `Service{ ensure => running }`. The unit is considered to be `is-active` (running for Puppet) although it exited

```
sudo systemctl status swapoff
● swapoff.service - Disable all swap partitions
   Loaded: loaded (/etc/systemd/system/swapoff.service; enabled)
   Active: active (exited) since Tue 2017-05-02 15:15:54 GMT; 11min ago
```

PR #1708, opened 2017-05-10
BO-174 Introduce fluentd plugin systemd to source events from journal

ppp0:


modules/fluentd/manifests/config/source_journald.pp
```diff
@@ -0,0 +1,35 @@
...
+      path /var/lib/fluentd/journald_pos
+    </storage>
+    | EOT
+
+
+    $config_template = @(EOC)
+    <source>
+      @type systemd
+      path <%= @path %>
+      tag <%= @fluentd_tag %>
```
njam: Will we know the source systemd-unit in Loggly?
Is it possible to send the unit as the tag?

ppp0: There is the possibility to "filter" the journal, see
https://github.com/reevoo/fluent-plugin-systemd/blob/31030ca07f688e7d705399063a79d8435efd1767/test/plugin/test_in_systemd.rb#L35
and then one would group the
```
<source>
    @type systemd
    tag unit-name
    filter [ { "_SYSTEMD_UNIT": "unit-name.service" } ]
</source>
```

but looking at the source code https://github.com/reevoo/fluent-plugin-systemd/blob/31030ca07f688e7d705399063a79d8435efd1767/lib/fluent/plugin/in_systemd.rb#L8 I fear this would result in a lot of single instantiations of this object?

Probably we're better off just removing all fields that we don't need from the journald record (quite a few) by using record-modifier, leave the field _SYSTEMD_UNIT - may be rename it - and then have loggly parse it?

ppp0:
Detail config of the whole logging to take place in puppet-cargomedia (future class `cargomedia_base::logging`)

ppp0:
retest this please

ppp0:
Following changes due to failing tests, adaptations to puppet-lint and puppet-syntax

modules/fluentd/manifests/config/source_journald.pp
```diff
@@ -0,0 +1,40 @@
+define fluentd::config::source_journald (
+  $path        = '/run/log/journal/',
+  $fluentd_tag = 'journal', # Can't be called "tag" in puppet
+  $priority    = 10,
+  $pos_file = true,
```
njam: |`line 5`|
|---|
Is there any reason to *not* have this? Otherwise, let's drop this parameter.

modules/fluentd/manifests/config/source_journald.pp
```diff
@@ -0,0 +1,40 @@
...
+    |- EOT
+
+
+    $config_template = @(EOC)
+    <source>
+      @type systemd
+      path <%= @path %>
+      tag <%= @fluentd_tag %>
+      read_from_head true
+      <%= @pos_file_template if @pos_file %>
```
njam: Also rewrite it according to the desired schema.

PR #1709, opened 2017-05-11
FB-115 Allow new Url pattern with env in path

zazabe:


modules/cm/manifests/vhost.pp
```diff
@@ -87,7 +87,7 @@
   }
 
   nginx::resource::location { "${name}-origin-upstream":
-    location            => '~* ^/(resources|vendor-css|vendor-js|library-css|library-js|layout)/',
+    location            => '~* ^.*/(resources|vendor-css|vendor-js|library-css|library-js|layout)/',
```
njam: I think we can just remove the `^` - which means "beginning of the string"?

PR #1713, opened 2017-05-24
BO-174 Add systemd support to fluentd

zazabe:


zazabe:
mmmh, all of this is a bit messy, some commits are not visible in this review, but rather in https://upsource.cargomedia.ch/puppet-packages/review/PP-CR-26, even if this one is closed already...
Anyhow, all commits are included in https://upsource.cargomedia.ch/puppet-packages/branch/PR%201713

modules/fluentd/spec/source_journald/spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  describe file('/var/lib/fluentd/journald_pos_systemd') do
+    it { should be_directory }
+  end
+
+  describe command('grep foo /tmp/dump/*.log | tail -1') do
+    its(:exit_status) { should eq 0 }
+    its(:stdout) do
+      is_expected.to include_json(
```
ppp0: Coool

ppp0:
spec test failure

ppp0:
The idea is to selectively enable journal source for fluentd with hiera
```
  "systemd::config::journald::fluentd_output": true,
```

due to the fact that we don't know yet much about the log volume to be expected

PR #1719, opened 2017-05-30
BO-56 Fluentd with journald support - bugfix

zazabe:


zazabe:
include changes from https://upsource.cargomedia.ch/puppet-packages/review/PP-CR-38 to avoid issue due to removal of `filter_record_modifier`

zazabe:
include changes from https://upsource.cargomedia.ch/puppet-packages/review/PP-CR-38 to avoid issue due to removal of `filter_record_modifier`

PR #1728, opened 2017-06-19
SYS-159 Remove log parsing on syslog

ppp0:


ppp0:
Might need manual purging of resources (and restart of bipbip) - At any rate, a supervised apply

PR #1733, opened 2017-06-20
SYS-59 Use uid email to check if an apt key is already added

zazabe:


modules/apt/manifests/key.pp
```diff
@@ -7,7 +7,7 @@
 ) {
   require 'apt'
 
-  $condition = "apt-key list | grep -E '^pub\\s+\\w+/${key}\\s+'"
+  $condition = "apt-key list | grep -E '^uid.+${key}'"
```
njam: $key is some fingerprint here, like "C857C906"

PR #1734, opened 2017-06-21
SYS-160 Mongodb logging to journal

ppp0:


zazabe:
it will restart mongo services, is it safe to merge as it is?

ppp0:
mergin on secondaries first, on master later (find out master via client, executing this: `printjson({db: db.isMaster()})`

PR #1737, opened 2017-06-27
FB-838 Add PHP timecop extension

dlondero:


modules/cm/manifests/application.pp
```diff
@@ -12,6 +12,7 @@
   require 'php5::extension::imagick'
   require 'php5::extension::curl'
   require 'php5::extension::runkit'
+  require 'php5::extension::timecop'
```
zazabe: Is it needed in prod? if not, i would require it only if `$development = true`

dlondero: That's just for tests so I agree to moving it only on dev!

PR #1750, opened 2017-07-04
SYS-198 Move replication status check from cron to bipbip

ppp0:


ppp0:
@{f4355e77-07da-4ca6-b033-d106f96ec3bb,Michel Meyer}I am not sure. OTOH it would be nice to say that replication fails *only and only if* the query result returns that replication it OFF. Al other cases (server is down, access to mysql is denied) should not assume anything. But this involves more complicated code as in `ff645da`

modules/mysql/manifests/server/slave.pp
```diff
@@ -23,11 +26,13 @@
...
   }
-  ->
 
-  cron { 'check-mysql':
-    command => '/usr/local/bin/mysql-replication-check',
-    user    => 'root',
+  @bipbip::entry { 'is-replication-running':
+    plugin  => 'command',
+    options => {
+      'command'      => "/usr/local/bin/mysql-replication-check root ${::mysql::server::root_password}",
```
zazabe: i understand we can't execute this cmd as root... but passing the mysql root password seems not very safe also... 
> SHOW STATUS provides server status information (see Section 5.1.7, “Server Status Variables”). This statement does not require any privilege. It requires only the ability to connect to the server. see https://dev.mysql.com/doc/refman/5.7/en/show-status.html

what about adding a `bipbip` user to mysql with not privilege?

ppp0:
fully agree

PR #1754, opened 2017-07-05
SYS-209 Add sudoers.d config to suppress all mail notification

ppp0:


modules/sudo/manifests/config/no_mail.pp
```diff
@@ -0,0 +1,15 @@
...
+  $mail_events = [
+    'mail_always',
+    'mail_badpass',
+    'mail_no_user',
+    'mail_no_host',
+    'mail_no_perms',
+  ]
+
+  @sudo::config { 'no-mail':
+    content => inline_template('<% @mail_events.each do |i| -%><%= "Defaults !#{i}\n" %><% end -%>'),
```
zazabe: I think the `on|off` flag is missing, it should be like:
```
Defaults mail_always off
Defaults mail_badpass off
...
```

ppp0:
No, the correct syntax is like it is: !<param> (some kind of boolean) - I tested it and it behaves as expected

zazabe:
fine

PR #1769, opened 2017-07-17
SYS-234 Log backup success/error to journal

zazabe:


modules/backup/templates/agent/rdiff/run.sh
```diff
@@ -0,0 +1,32 @@
...
+      message="${message}, see ${OUTPUT_FILE}"
+    fi
+    log_error "${message}"
+  fi
+}
+
+trap catch_error ERR
+
+OUTPUT=$(/usr/local/bin/backup-${ACTION}.sh "$@" 2>&1)
+log_info "mysql backup ${ACTION} done successfully"
```
njam: It's not only for "mysql", see $sourceType variable

modules/backup/templates/agent/rdiff/run.sh
```diff
@@ -0,0 +1,32 @@
...
+  logger -p 3 "$1"
+}
+
+catch_error () {
+  EXIT_CODE="${?}"
+  if (test ${EXIT_CODE} -gt 0); then
+    message="mysql.backup-${ACTION} failed with exit code ${EXIT_CODE}"
+
+    if [ "${OUTPUT}" != "" ]; then
+      OUTPUT_FILE="/tmp/mysql.backup-${ACTION}.$(date '+%s').out"
```
njam: /var/log might be better so it doesn't get lost on reboot or other actions?

modules/backup/templates/agent/rdiff/run.sh
zazabe: right, i thought about making it more generic, a trap for any cron job basically

zazabe:
@njam please check again

modules/backup/manifests/agent.pp
```diff
@@ -22,14 +22,14 @@
   }
 
   cron { "backup-${name}":
-    command => "/usr/local/bin/backup-create.sh -h '${host}' -s '${source}' -d '${destination}' -o '${options}' -t '${sourceType}' -r '${remove_after}'",
+    command => "/usr/local/bin/backup-run.sh ${sourceType} create -h '${host}' -s '${source}' -d '${destination}' -o '${options}' -t '${sourceType}' -r '${remove_after}'",
```
njam: Should we then remove "-t" here and pass it from `backup-run.sh`? Not sure.

modules/backup/templates/agent/rdiff/run.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -e
```
njam: |`line 8`|
|---|
nit: indentation

modules/backup/templates/agent/rdiff/run.sh
```diff
@@ -0,0 +1,33 @@
...
+
+TYPE="${1}" ; shift
+ACTION="${1}" ; shift
+OUTPUT=""
+
+log_info () {
```
njam: nit: since you use `if (test ...` maybe also use it here instead of `if [ ...` - or the other way around?

modules/backup/templates/agent/rdiff/run.sh
```diff
@@ -0,0 +1,33 @@
...
+
+log_error () {
+  logger -p 3 "$1"
+}
+
+catch_error () {
+  EXIT_CODE="${?}"
+  if (test ${EXIT_CODE} -gt 0); then
+    message="backup ${ACTION}:${TYPE} failed with exit code ${EXIT_CODE}"
+
```
njam: Very nice script!

modules/backup/manifests/agent.pp
zazabe: hum, this first arg is here just to identify the type of backup, but it could be different than `-t`, also the `backup-run.sh` will have to understand subcommand arguments...

PR #1779, opened 2017-07-20
SYS-235 Set jetbrains services config with command-line API

zazabe:


modules/jetbrains/manifests/application.pp
```diff
@@ -23,6 +23,10 @@
   $config_path = "${home_path}/conf"
   $var_path = "/var/lib/${service_name}"
 
+  $config_file = "${config_path}/internal/bundle.properties"
+  $config_command = "/usr/local/${service_name}/bin/${name}.sh"
+  $config_arguments = jetbrains_configure_arguments(inline_template($config))
+
   $installation_uuid = fqdn_uuid("${name}.${host}")
```
njam: This way puppet will not apply changes to the config, but only configure the application once in the beginning.

How about instead defining multiple `exec` resources which each set a single config value, *if* it is not already set?
Something along the lines of:
```
  exec { "configure ${service_name} ${property_name}":
    provider => shell,
    command  => "/usr/local/jetbrains-youtrack/bin/youtrack.sh configure --${property_name}=${property_value}",
    path     => ['/usr/sbin', '/usr/bin', '/sbin', '/bin'],
    unless   => "grep ${property_name}=${property_value} ${config_file}",
  }
```

But it also feels a bit fiddly. Maybe it would be better to use augeas in the end?

zazabe:
I did it like that at first, with a `create_resources('exec', ...)`, but it doesn't really help i think.

I prefer the script, because augeas only works if the config file already exists, which is not the case when we install/upgrade a service. 
On the other hand the script will apply the config in his own way (I presume it's not only about writing values in a file), with the correct escaping, etc...

Moreover, the common use-case is to upgrade, and in this case it's equivalent to a re-installation + backup restore. 

I agree that with this changes, we can't really change the config of a running version.

It could be achieved by using multiple exec, stopping the app if there's some changes, running the script for each change, and then restarting the app, but it seems a bit complex for something that will be very rarely used.

njam:
Agree, let's keep it like this, we just need to remember about it.

PR #1780, opened 2017-07-20
SYS-232 Upgrade nodejs to 6.x

zazabe:


modules/nodejs/manifests/init.pp
```diff
@@ -14,12 +14,13 @@
       }
     },
     require => Class['apt::transport_https'],
+    notify  => Package['nodejs'],
   }
-  ->
 
   package { 'nodejs':
-    ensure   => present,
+    ensure   => 'latest',
```
njam: I assume this repo will get security updates, which means puppet will upgrade the package automatically for us. Probably we don't want that, because it would restart critical services.
Shouldn't we keep "present" and upgrade manually using pulsar?

zazabe:
agree

btw, it will really restart all services depending on node? I was wondering if we should not add a `subscribe` between those services and `Package['nodejs']`

njam:
I thought we had it, but might not be true.
Either way "present" is safer.

PR #1787, opened 2017-07-27
SYS-225 Add resource constraints to ImageMagick

ppp0:


ppp0:
Please check https://www.imagemagick.org/script/resources.php for policy.xml - Be aware that our version 6.8.9 does not support the resources "width" and "height"

ppp0:
The rationale for setting these values: 
* Max size as in https://github.com/cargomedia/sk/blob/master/library/SK/library/SK/FormField/PhotoUpload.php#L5
* Calculation done with http://www.scantips.com/basics1d.html (using 16-bit color depth * 3 colors RGB = 48-bit color depth)
* As much as possible, RAM should be used for performance purposes
* Disk should be used as little as possible - and limited

PR #1816, opened 2017-09-20
FB-1340 Add fluentd elasticsearch plugin adn config

tomaszdurka:


modules/fluentd/manifests/config/match_elasticsearch.pp
```diff
@@ -0,0 +1,32 @@
+define fluentd::config::match_elasticsearch (
+  $pattern,
+  $priority        = 80,
+  $hosts,
+  $index_name,
+  $type_name,
+  $include_tag_key = true,
+  $logstash_format = false,
```
njam: how about making it TRUE?

njam:
The build is failing, but doesn't list any failed examples. Can you check it?

tomaszdurka:
So it seems the usage of "intended-heredoc" (<<~HEREDOC) was a problem. It is introduced in ruby 2.3. 
Locally it worked fine as I am using 2.4 and in the jenkins environment it is lower.

modules/fluentd/manifests/config/match_elasticsearch.pp
tomaszdurka: I thought that if it's named elasticsearch (not logstash) we should treat it as option. As it also ignores index_name field. I set it to true in cargomedia::out::elasticsearch though.

tomaszdurka:
I will go ahead and merge to work on cargomedia module. We can always add follow up.

PR #1822, opened 2017-10-31
FB-1492 Add proxy_next_upstream conditions, set timeout

ppp0:


ppp0:
Great! Adding `http_503` on VM made the proxy pass a request when it received a maintenance page from upstream!

njam:
My understanding is that if a request is received, nginx will send it to the first upstream, and if that fails it will try the next one ''if'' less than 5 seconds have passed. So this will affect cases where the request took a long time to fail. What are such cases? Maybe that the server is kind of "stuck"? In such a case wouldn't it be desirable to try the next server?

ppp0:
As it is now, the default of 'timeout 0' is set (indefinite wait) - I'd rather not set timeout at all in that case. The nodes themselves have their own fpm timeout (300 sec?)

PR #1823, opened 2017-11-08
FB-1490 Use git:// scheme for Github Puppetfile-sources

njam:
This way librarian-puppet uses HTTP to connect to Github, so we don't need an SSH key.


njam:
The idea is to use HTTP cloning when we access modules from the public repo. This way we don't need an SSH key in many cases which should simplify provisioning with puppet.

njam:
Build failures are unrelated to this change.

PR #1824, opened 2017-11-08
FB-1555 Introduce network::route resource, declutter, remove obsoleted resources

ppp0:


modules/network/manifests/init.pp
```diff
@@ -1,12 +1,11 @@
...
-  require 'network::interfaces_backup'
-
-  $interfaces = lookup('network::interfaces', Hash, 'deep', { })
-  create_resources('network::interface', $interfaces)
-
   $hosts = lookup('network::hosts', Hash, 'deep', { })
   create_resources('network::host', $hosts)
 
+  $routes = lookup('network::routes', Hash, 'deep', { })
+  create_resources('@network::route', $routes)
```
njam: Why creating a _virtual_ resource here, and then realizing it 2 lines later? Wouldn't it be the same as just realizing it (as it was done for interfaces)?

modules/network/manifests/route.pp
```diff
@@ -0,0 +1,22 @@
...
+  # This file is only executed at boot time
+  file { "/etc/network/if-up.d/99-${title}":
+    ensure  => file,
+    content => template("${module_name}/route/if-up.erb"),
+    owner   => '0',
+    group   => '0',
+    mode    => '0750',
+  }
+
+  exec { "[${title}] Set route to ${destination} via ${via}":
```
njam: How about executing the script created above here, to avoid duplication?
The IF could also be moved into such script.

zazabe:
@{c7e8ab9c-90bf-4796-bb88-9aa7c8150044,Reto Kaiser} please check again

modules/network/templates/route/if-up.erb
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+if ! /sbin/ip ro sh | grep -v -q '${destination} via ${via}'
+then
```
njam: I would use the long and documented form everywhere, so `ip route list` instead of `ip ro sh`.

njam:
Specs fail

njam:
We don't have a spec test for destination="default gw" which we actually use for production.
Could check if we can add a test, and if not make sure manually it works fine?

njam:
Please also check in that case that the resource doesn't apply on a second run. I have some concerns about the IF we're using.

modules/network/manifests/route.pp
```diff
@@ -0,0 +1,23 @@
+define network::route (
+  $destination = 'default gw',
+  $via,
+)
+{
+  # This file is only executed at boot time
```
njam: We could also remove the `unless`, add `refreshonly => true` and notify this resource from the `exec`.

PR #1833, opened 2017-12-06
Remove revive maintenance cron task

zazabe:


njam:
Did you want to remove the backup?
That's configured here: https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/files/hiera-data/datacenter/softlayer/vlan1850-metal/ads4.json#L10

PR #1845, opened 2018-01-09
Revert "Adjust GPG signature key for mongodb deb repo"

ppp0:
Reverts cargomedia/puppet-packages#1844


ppp0:
This is needed to install without errors. Will be re-reverted later

PR #1847, opened 2018-01-25
Convert fpm scripts to prometheus format

tomaszdurka:


modules/php5/templates/extension/opcache/opcache-status.php
```diff
@@ -1,4 +1,16 @@
...
 
 $status = opcache_get_status(false);
-echo json_encode($status);
+$metrics = new RecursiveIteratorIterator(new RecursiveArrayIterator($status));
+$prefix = 'php_opcache';
+
+foreach ($metrics as $name => $value) {
+    $help = ucfirst(str_replace('_', ' ', $name));
+    echo "# HELP {$prefix}_{$name} {$help}\n";
+    echo "# TYPE {$prefix}_{$name} gauge\n";
```
njam: some are counter

see https://github.com/cargomedia/bipbip/blob/master/lib/bipbip/plugin/fastcgi_php_opcache.rb

